PR #42, opened 2024-03-25
feat: add new build_clients logic

### LINKED ISSUE #58: `loam build` also builds clients from environments.toml settings
This new functionality should behave as described in https://github.com/loambuild/frontend?tab=readme-ov-file#loam-dev

chadoh:
Fixes #58 

At the end of a `loam build` run, it will now run `build_clients` logic. This looks for an `enviroments.toml` file like the kind shown [in loambuild/frontend](https://github.com/loambuild/frontend/blob/main/environments.toml), and automatically gets a frontend to a point where it would be ready to run. For example, running with the `environments.toml` file shown in the new `tests/it/build_clients/contracts.rs` file here, you get this output:

```
$ loam build
   Compiling loam-cli v0.8.0 (/Users/chadoh/code/loam/sdk/crates/loam-cli)
    Finished dev [unoptimized + debuginfo] target(s) in 5.43s
     Running `target/debug/loam build --manifest-path ./crates/loam-cli/tests/fixtures/soroban-init-boilerplate/Cargo.toml`
cargo rustc --manifest-path=crates/loam-cli/tests/fixtures/soroban-init-boilerplate/contracts/auth/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
    Finished release [optimized] target(s) in 0.07s
cargo rustc --manifest-path=crates/loam-cli/tests/fixtures/soroban-init-boilerplate/contracts/hello_world/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
    Finished release [optimized] target(s) in 0.04s
cargo rustc --manifest-path=crates/loam-cli/tests/fixtures/soroban-init-boilerplate/contracts/increment/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
    Finished release [optimized] target(s) in 0.04s
cargo rustc --manifest-path=crates/loam-cli/tests/fixtures/soroban-init-boilerplate/contracts/custom_types/Cargo.toml --crate-type=cdylib --target=wasm32-unknown-unknown --release
    Finished release [optimized] target(s) in 0.04s
🌐 using network at http://localhost:8000/rpc
🔐 creating keys for "alice"
📲 installing "hello_world" wasm bytecode on-chain...
    ↳ hash: 97183d1986ce664266ba148e98a472b58064032cb7b57f2e2311ec2311944892
🪞 instantiating "hello_world" smart contract
    ↳ contract_id: CBTZIBW5MHZAXSRJELN22TNP7G7JD4WTN46652T5EAFLFKA6XR3GMJSO
🎭 binding "hello_world" contract

added 3 packages, and audited 47 packages in 621ms

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> hello_world@0.0.0 build
> tsc

🍽️ importing "hello_world" contract
📲 installing "soroban_auth_contract" wasm bytecode on-chain...
    ↳ hash: 2860c7f557f065d80d2e26e2e49146c6ced5f5b84f106d5c2e1733206eab3cb4
🪞 instantiating "soroban_auth_contract" smart contract
    ↳ contract_id: CA7PQX6T5QLV73PI27IANYYVINWYNKP7ZUEKCPOMDSZFHO5NLOODQYC5
🎭 binding "soroban_auth_contract" contract

added 3 packages, and audited 47 packages in 547ms

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> soroban_auth_contract@0.0.0 build
> tsc

🍽️ importing "soroban_auth_contract" contract
📲 installing "soroban_custom_types_contract" wasm bytecode on-chain...
    ↳ hash: 57c7d3125a06aec612ad2f9e89ee44c6095827b9b78d15b16bd4aa94dc8b2341
🪞 instantiating "soroban_custom_types_contract" smart contract
    ↳ contract_id: CDNR257FK3IDXDEQP4ESQC4BPXO2LGSCAWOQ2DMWOYEGMLYA7U7RLKFR
🎭 binding "soroban_custom_types_contract" contract

added 3 packages, and audited 47 packages in 529ms

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> soroban_custom_types_contract@0.0.0 build
> tsc

🍽️ importing "soroban_custom_types_contract" contract
📲 installing "soroban_increment_contract" wasm bytecode on-chain...
    ↳ hash: 7245f2e1717756379749da3c65907534808b44ccbf90d53c9b380ecb27bdaa72
🪞 instantiating "soroban_increment_contract" smart contract
    ↳ contract_id: CBQUBFAXPNXNBU7Q75MVPFUCSMD3T47DBFBISB6MO7UYQCPZK5VICBQL
🎭 binding "soroban_increment_contract" contract

added 3 packages, and audited 47 packages in 552ms

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> soroban_increment_contract@0.0.0 build
> tsc

🍽️ importing "soroban_increment_contract" contract
```

(we might want to consider [suppressing the stdout](https://users.rust-lang.org/t/silencing-subprocess-output/46626/3) output of `soroban contract bindings typescript` by default)

## Still simplistic

This is a good-enough starting point, and I'd like to merge it now so other people can help with some improvements. Right now, this logic is no better than [the `initialize.js` script in soroban-astro-template](https://github.com/stellar/soroban-astro-template/blob/main/initialize.js). Notably, it does not:

- #72
- #73
- #74 
- #75
- #76
- #77


.cargo/config.toml
```diff
@@ -1,7 +1,7 @@
 # paths = ["/path/to/override"] # path dependency overrides
 
 [alias] # command aliases
-
+install_soroban = "install --version 21.0.0-rc.1 --root ./target soroban-cli --debug"
```
chadoh, line 4: @willemneal is there a way to do something like [the js-stellar-sdk `test/e2e/initialize.sh` script](https://github.com/stellar/js-stellar-sdk/blob/840f57da0d2fe3cc52d37670becd76ecd8b2cdca/test/e2e/initialize.sh#L17-L30) for our integration tests? Before the integration tests run, I would like to make sure that the correct version of soroban is pinned, and maybe also run the `soroban contract init` command that's currently being used to build the `soroban-init-boilerplate` (see below).

crates/loam-cli/tests/it/build_clients/contracts.rs
```diff
@@ -0,0 +1,96 @@
...
+
+        env.loam("build").assert().success().stdout(
+            predicates::str::contains("creating keys for \"alice\"\n")
+                .and(predicates::str::contains(
+                    "using network at http://localhost:8000/rpc\n",
+                ))
+                .and(contract_stdout_predicates.pop().unwrap())
+                .and(contract_stdout_predicates.pop().unwrap())
+                .and(contract_stdout_predicates.pop().unwrap())
+                .and(contract_stdout_predicates.pop().unwrap()),
```
chadoh, line 60: @willemneal as you might guess from this code, I would like to use `fold` above, instead of `map`. The first thing to try is writing it like this:

```rs
let mut contract_stdout_predicates = contracts
    .iter()
    .fold(predicates::str::contains(""), |acc, c| {
        acc.and(predicates::str::contains(format!(
            "installing \"{c}\" wasm bytecode on-chain"
        )))
        .and(predicates::str::contains(format!(
            "instantiating \"{c}\" smart contract"
        )))
        .and(predicates::str::contains(format!(
            "binding \"{c}\" contract"
        )))
        .and(predicates::str::contains(format!(
            "importing \"{c}\" contract"
        )))
    })
    .collect::<Vec<_>>();
```

But this gives the following error:

```
mismatched types
expected struct 'ContainsPredicate
found struct
'AndPredicate<AndPredicate<AndPredicate<AndPredicate<ContainsPredicate, ContainsPredicate, str>, ContainsPredicate, str>, ContainsPredicate, str>,
ContainsPredicate, str>' (rustc E0308)
```

So basically, Rust doesn't like that the first run of the loop will have a `ContainsPredicate` but subsequent iterations will have some number of `AndPredicate`s piled on there. But the `.stdout(...)` matcher doesn't care, so how do we tell `fold` that it's ok?

I tried this:

```rs
let mut contract_stdout_predicates = contracts.iter().fold(
    predicates::str::contains("") as IntoOutputPredicate<str>,
    |acc, c| {
        acc.and(predicates::str::contains(format!(
            "installing \"{c}\" wasm bytecode on-chain"
        )))
        .and(predicates::str::contains(format!(
            "instantiating \"{c}\" smart contract"
        )))
        .and(predicates::str::contains(format!(
            "binding \"{c}\" contract"
        )))
        .and(predicates::str::contains(format!(
            "importing \"{c}\" contract"
        )))
    },
);
```

But now I get an even scarier error:

```
 cast to unsized type: `ContainsPredicate` as `dyn IntoOutputPredicate<str>`
```

Any other ideas of things to try?

crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -1,23 +1,23 @@
 #![allow(clippy::struct_excessive_bools)]
+use crate::commands::build::environments_toml;
```
willemneal, line 2: Could this be `envs_toml`? Seems long if you are going to use it as a prefix.

crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -102,13 +55,17 @@ impl Cmd {
         Ok(())
     }
 
+    fn loam_env(&self) -> &str {
```
willemneal, line 58: See other comment about Enum. You can add `#[default]` above a case in the enum that is the default. Or you can provide the default value in the clap args above the field.

.cargo/config.toml
chadoh: That looks great! Added an issue:

- https://github.com/loambuild/loam-sdk/issues/68

For this PR, I noticed that we already have `just setup` that installs soroban-cli, and I updated that version to match the one that this PR relies on.

crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -1,23 +1,23 @@
...
-use std::ops::Deref;
-use std::{fmt::Debug, io};
+use std::fmt::Debug;
 
 #[derive(Parser, Debug, Clone)]
 pub struct Cmd {
     #[arg(long, default_value = ".")]
     pub workspace_root: std::path::PathBuf,
+    #[arg(env = "LOAM_ENV", default_value = "production")]
+    pub env: Option<String>,
```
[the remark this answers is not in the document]
chadoh: Good call, thanks!

crates/loam-cli/src/commands/build/build_clients.rs
chadoh: Sure, even better, let's drop that `s`. Renamed to `env_toml`

crates/loam-cli/src/commands/build/build_clients.rs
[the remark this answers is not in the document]
chadoh: Actually, the `Environments` struct currently allows the keys in this file to be _anything_, which maybe is fine? If you want to have a `lol` environment, who are we to stop you.

Not sure it's worth changing it to match rn, though

crates/loam-cli/tests/it/build_clients/contracts.rs
willemneal: I would first not collect back to a vec but keep it as an iterator. Then you can use it in a for loop.
```rust
let mut res = env.loam("build").assert().success();
for predicate in &contract_stdout_predicates {
    res = res.stdout(predicate);
}
```
Your fighting against the iterator. Let it flow!

willemneal: You might need to make `res` be `&mut Assert`. Or make a new variable with the return type of `.stdout`

willemneal: Or if you wanted to use fold, I would use the Assert. Since `.stdout` can be called each time and not one huge predicate.

crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -0,0 +1,260 @@
...
+                    .then(|| account.name.clone())
+            })
+            .collect::<Vec<_>>();
+
+        let default_account = match default_account_candidates.len() {
+            0 if accounts.is_empty() => return Err(Error::NeedAtLeastOneAccount),
+            0 => accounts[0].name.clone(),
+            1 => default_account_candidates[0].to_string(),
+            _ => return Err(Error::OnlyOneDefaultAccount(default_account_candidates)),
+        };
```
willemneal, line 134:
```suggestion
        let default_account = match (default_account_candidates, accounts) {
            ([], []) => return Err(Error::NeedAtLeastOneAccount),
            ([], [Account {name, ..}, ..] => name.clone(),
            ([candidate], _) => candidate.to_string(),
            _ => return Err(Error::OnlyOneDefaultAccount(default_account_candidates)),
        };
```

willemneal: Also do you need `candidate.to_string`

chadoh: Your suggestion doesn't work for a bunch of reasons and my Rust is not good enough to get it to work. Something to do with converting vecs to slices and then matching with arrays and then `the size for values of type `[Account]` cannot be known at compilation time`.

chadoh: Yes, we need to the `to_string` to work around 

```
cannot move out of index of `Vec<std::string::String>`
move occurs because value has type `std::string::String`, which does not
implement the `Copy` trait (rustc E0507)

Comment (chadoh) [reply crates/loam-cli/src/commands/build/build_clients.rs:134]: If you want to edit the code yourself and push up a change, I'd love to see how to do that nicer `match`

File: crates/loam-cli/Cargo.toml [hunk 7]
```diff
@@ -51,8 +51,11 @@ heck = "0.5.0"
 pathdiff = "0.2.1"
 shlex = "1.1.0"
 symlink = "0.1.0"
+toml = { version = "0.8.12", features = ["parse"] }
+soroban-cli = { git = "https://github.com/stellar/soroban-cli.git", rev = "e189c710f8eee112278a059bccb80a9eb6f9b49f" }
```
willemneal, line 55: Can you use the newest release?

crates/loam-cli/src/commands/build/env_toml.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[derive(Debug, serde::Deserialize, Clone)]
+pub struct Account {
+    pub name: String,
+    pub default: Option<bool>,
+}
+
+#[derive(Debug, serde::Deserialize, Clone)]
+pub struct Contract {
+    pub workspace: Option<bool>,
```
willemneal, line 49: Same here.

crates/loam-cli/src/commands/build/env_toml.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn get(workspace_root: &PathBuf, loam_env: String) -> Result<Option<Environment>, Error> {
+        let env_toml = workspace_root.join("environments.toml");
+
+        if !env_toml.exists() {
+            return Ok(None);
+        }
+
+        let toml_str = std::fs::read_to_string(env_toml).map_err(Error::ParsingToml)?;
+        let parsed_toml: Environments = toml::from_str(&toml_str).unwrap();
+        let current_env = parsed_toml.get(loam_env.as_str());
```
willemneal, line 62:
```suggestion
        let current_env = parsed_toml.get(loam_env);
```

crates/loam-cli/src/commands/build/env_toml.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub default: Option<bool>,
+}
+
+#[derive(Debug, serde::Deserialize, Clone)]
+pub struct Contract {
+    pub workspace: Option<bool>,
+}
+
+impl Environment {
+    pub fn get(workspace_root: &PathBuf, loam_env: String) -> Result<Option<Environment>, Error> {
```
willemneal, line 53:
```suggestion
    pub fn get(workspace_root: &Path, loam_env: &str) -> Result<Option<Environment>, Error> {
```

crates/loam-cli/Cargo.toml
chadoh: Good call, done

crates/loam-cli/src/commands/build/env_toml.rs
```diff
@@ -0,0 +1,68 @@
...
+            return Ok(None);
+        }
+
+        let toml_str = std::fs::read_to_string(env_toml).map_err(Error::ParsingToml)?;
+        let parsed_toml: Environments = toml::from_str(&toml_str).unwrap();
+        let current_env = parsed_toml.get(loam_env.as_str());
+        if current_env.is_none() {
+            return Err(Error::NoSettingsForCurrentEnv(loam_env));
+        };
+        Ok(current_env.cloned())
```
willemneal, line 66: Why clone?

crates/loam-cli/src/commands/build/env_toml.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        if !env_toml.exists() {
+            return Ok(None);
+        }
+
+        let toml_str = std::fs::read_to_string(env_toml).map_err(Error::ParsingToml)?;
+        let parsed_toml: Environments = toml::from_str(&toml_str).unwrap();
+        let current_env = parsed_toml.get(loam_env.as_str());
+        if current_env.is_none() {
+            return Err(Error::NoSettingsForCurrentEnv(loam_env));
```
willemneal, line 64:
```suggestion
            return Err(Error::NoSettingsForCurrentEnv(loam_env.to_owned()));
```

crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -0,0 +1,229 @@
...
+        format!("{:?}", self.env).to_lowercase()
+    }
+
+    /// Parse the network settings from the environments.toml file and set STELLAR_RPC_URL and
+    /// STELLAR_NETWORK_PASSPHRASE.
+    ///
+    /// We could set STELLAR_NETWORK instead, but when importing contracts, we want to hard-code
+    /// the network passphrase. So if given a network name, we use soroban-cli to fetch the RPC url
+    /// & passphrase for that named network, and still set the environment variables.
+    fn add_network_to_env(&self, network: &env_toml::Network) -> Result<(), Error> {
```
willemneal, line 75: Why is this a instance method? `self` isn't referenced at all.

crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -0,0 +1,229 @@
...
+            } = (cli::network::Args {
+                network: Some(name.clone()),
+                rpc_url: None,
+                network_passphrase: None,
+            })
+            .get(&cli::config::locator::Args {
+                global: false,
+                config_dir: None,
+            })?;
+            println!("🌐 using {name} network");
```
willemneal, line 93:
```suggestion
            eprintln!("🌐 using {name} network");
```
Only program outputs that are important to be consumed should go to stdout.

crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    /// Parse the network settings from the environments.toml file and set STELLAR_RPC_URL and
+    /// STELLAR_NETWORK_PASSPHRASE.
+    ///
+    /// We could set STELLAR_NETWORK instead, but when importing contracts, we want to hard-code
+    /// the network passphrase. So if given a network name, we use soroban-cli to fetch the RPC url
+    /// & passphrase for that named network, and still set the environment variables.
+    fn add_network_to_env(&self, network: &env_toml::Network) -> Result<(), Error> {
+        let rpc_url = &network.rpc_url;
+        let network_passphrase = &network.network_passphrase;
```
willemneal, line 77: You could do a match here on `&network`. It would  be a bit cleaner, especially the nested if lets. e.g.

```rust
match network {
    Network { name: Some(name), ..}  => ...
    Network { rpc_url: Some(rpc_url), network_passphrase: Some(network_passphrase, ..} => ...
    _ => return Err
}
```

crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -0,0 +1,229 @@
...
+                println!("🌐 using network at {rpc_url}");
+            } else {
+                return Err(Error::MalformedNetwork);
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn handle_accounts(
```
willemneal, line 109: same here with self

crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -0,0 +1,229 @@
...
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, clap::ValueEnum)]
+pub enum LoamEnv {
+    Development,
+    Testing,
+    Staging,
+    Production,
+}
+
+#[derive(Parser, Debug, Clone)]
+pub struct Cmd {
```
willemneal, line 18: Is this actually used as a command? Perhaps you could turn this into an Args with workspace_root as an option. Then you could flatten it into the build command.

willemneal (changes requested):
Couple of design questions.

crates/loam-cli/src/commands/build/env_toml.rs
```diff
@@ -0,0 +1,68 @@
...
+#[derive(Debug, serde::Deserialize, Clone)]
+pub struct Environment {
+    pub accounts: Option<Vec<Account>>,
+    pub network: Network,
+    pub contracts: Option<Map<Box<str>, Contract>>,
+}
+
+#[derive(Debug, serde::Deserialize, Clone)]
+#[serde(rename_all = "kebab-case")]
+pub struct Network {
```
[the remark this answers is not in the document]
chadoh: I think we should keep our own definition. Ours is describing the format of the `environments.toml` file, which already differs slightly from the definition in stellar-cli (`name` vs `network`), and which may deviate more in the future. Additionally, the one in stellar-cli does not derive `serde::Deserialize`, and there's not really a reason for it to do so.

crates/loam-cli/src/commands/build/env_toml.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub name: Option<String>,
+    pub rpc_url: Option<String>,
+    pub network_passphrase: Option<String>,
+    // run_locally: Option<bool>,
+}
+
+#[derive(Debug, serde::Deserialize, Clone)]
+pub struct Account {
+    pub name: String,
+    pub default: Option<bool>,
```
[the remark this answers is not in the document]
chadoh: We only really need the `skip_serializing_if` if we plan to _serialize_ it, right? Right now we never serialize these, only deserialize.

crates/loam-cli/src/commands/build/env_toml.rs
chadoh: Could do this instead:

```rs
        let current_env = parsed_toml.get(&*loam_env);
```

Is that preferable to `loam_env.as_str()`?

crates/loam-cli/src/commands/build/env_toml.rs
chadoh: oh, I see, you were suggesting combining this with the above suggestion. sure.

crates/loam-cli/src/commands/build/env_toml.rs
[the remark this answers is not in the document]
chadoh: If I remove the `Option` we get deserialization errors

    TomlError { message: "missing field `default`"

crates/loam-cli/src/commands/build/env_toml.rs
chadoh: Because `Map#get` returns an option of a reference to the thing, and we need to return the thing itself (or a clone of it)

chadoh:
All addressed, thank youuuuuu

willemneal (approved):

PR #66, opened 2024-06-11
chore: update readme.md

gfarrenkopf:
link numbered list to github repos

### DIFF
--- a/crates/loam-sdk/README.md
+++ b/crates/loam-sdk/README.md
@@ -1,6 +1,10 @@
 # loam-sdk
 
-Build **composable, upgradeable, secure Smart Contracts.
+Build **composable**, **upgradeable**, **secure** Smart Contracts.
+
+- **Composable**: With Loam SDK, you compose your smart contract from many _sub_ contracts. Subcontracts are like lego blocks that you can either use off-the-shelf from the open source ecosystem or that you can build yourself. A single Loam _smart_ contract is composed of one or more subcontracts.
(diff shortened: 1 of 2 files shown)


crates/loam-sdk/README.md
```diff
@@ -1,5 +1,7 @@
 # loam-sdk
 
+Build **composable, upgradeable, secure Smart Contracts.
+
```
chadoh, line 4:
```suggestion
Build **composable**, **upgradeable**, **secure** Smart Contracts.

- **Composable**: With Loam SDK, you compose your smart contract from many _sub_ contracts. Subcontracts are like lego blocks that you can either use off-the-shelf from the open source ecosystem or that you can build yourself. A single Loam _smart_ contract is composed of one or more subcontracts.
- **Upgradeable**: The one subcontract that all Loam smart contracts must include ([loam-subcontract-core](../loam-subcontract-core)) adds an important method to the smart contract: `redeploy`. You can call this method to switch the `wasm` hash—the behavior/brains of the contract—to a new one, while keeping the same contract ID. The storage accessed by each particular subcontract is loaded lazily, so upgrading one subcontract does not require migrating the data of another; each subcontract within your smart contract can be considered and upgraded independently.
- **Secure**: The [core subcontract](../loam-subcontract-core) also adds `admin_set` and `admin_get` to your contract, to make sure that only your trusted admin account can call `redeploy`. Our full loam architecture, beyond Loam SDK, also includes a universal factory contract, which makes it possible to deploy your contract and call `admin_set` in a single transaction, helping avoid front-running.

```

willemneal (approved):

PR #71, opened 2024-06-14
Add binaries to releases

elizabethengelman:
Closes: https://github.com/loambuild/loam-sdk/issues/53

- [x] Automatically build binaries and add them to releases when a new release is published
- [x] Create an npm package that wraps the rust binary
- [x] Publish an updated npm version via a gh action workflow

* I tested the workflows in my fork, and they seemed to work over there. 
* I messed up the npm version while testing and "unpublished" v0.1.1, so when we publish again we'll need to jump to v0.1.2 😬


.github/workflows/cd.yml
```diff
@@ -24,7 +24,7 @@ defaults:
 jobs:
```
elizabethengelman, line 24: The changes in this file allow us to build the binaries and add them to the most recent release. This workflow is automatically kicked off when a new release is created.

.github/workflows/npm.yml
```diff
@@ -0,0 +1,33 @@
+name: Build and publish npm package
+
+on:
+  workflow_dispatch:
```
elizabethengelman, line 4: I made this a manual workflow for now, because I wasn't sure exactly what we'd want the process for bumping the version in package.json to look like. It seems like maybe that version number should only be bumped when the loam-cli's version is released. Curious about what others think!

.github/workflows/npm.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: npm install
+        working-directory: npm/
+
+      - name: Build the project
+        run: npm run build
+        working-directory: npm/
+
+      - name: Publish package
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
elizabethengelman, line 31: This package is currently in my personal npmjs account. Happy to add other collaborators/move it to a loam org instead.

.github/workflows/npm.yml
willemneal: Yeah in the past I found it easiest to keep the versions in sync. I think for now we can go with manual just to start, but ideally it would be automatic.

willemneal (approved):
Let's give it a try!

.github/workflows/npm.yml
willemneal: Definitely! But let's try with yours for now.  I'm sure if you have the permissions but if you dm the secret I can add it.

PR #87, opened 2024-06-24
add dev command

BlaineHeffron:
adds dev command

### DIFF
--- a/crates/loam-cli/src/commands/build/clients.rs
+++ b/crates/loam-cli/src/commands/build/clients.rs
@@ -70,7 +70,7 @@ impl Args {
     }
 
     fn loam_env(self, default: LoamEnv) -> String {
-        self.env.unwrap_or(default).to_string()
+        self.env.unwrap_or(default).to_lowercase()
     }
 
     /// Parse the network settings from the environments.toml file and set `STELLAR_RPC_URL` and
--- a/crates/loam-cli/src/commands/build/clients.rs
+++ b/crates/loam-cli/src/commands/build/clients.rs
@@ -70,7 +70,7 @@ impl Args {
     }
 
     fn loam_env(self, default: LoamEnv) -> String {
-        self.env.unwrap_or(default).to_lowercase()
+        self.env.unwrap_or(default).to_string().to_lowercase()
     }
 
     /// Parse the network settings from the environments.toml file and set `STELLAR_RPC_URL` and
--- a/crates/loam-cli/src/commands/build/clients.rs
+++ b/crates/loam-cli/src/commands/build/clients.rs
@@ -15,10 +15,16 @@ pub enum LoamEnv {
     Production,
 }
 
+impl std::fmt::Display for LoamEnv {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", format!("{self:?}").to_lowercase())
+    }
+}
+
 #[derive(clap::Args, Debug, Clone, Copy)]
 pub struct Args {
-    #[arg(env = "LOAM_ENV", value_enum, default_value = "production")]
-    pub env: LoamEnv,
+    #[arg(env = "LOAM_ENV", value_enum)]
+    pub env: Option<LoamEnv>,
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -49,7 +55,8 @@ pub enum Error {
 
 impl Args {
     pub async fn run(&self, workspace_root: &std::path::Path) -> Result<(), Error> {
-        let Some(current_env) = env_toml::Environment::get(workspace_root, &self.loam_env())?
+        let Some(current_env) =
+            env_toml::Environment::get(workspace_root, &self.loam_env(LoamEnv::Production))?
         else {
             return Ok(());
         };
@@ -62,8 +69,8 @@ impl Args {
         Ok(())
     }
 
-    fn loam_env(self) -> String {
-        format!("{:?}", self.env).to_lowercase()
+    fn loam_env(self, default: LoamEnv) -> String {
+        self.env.unwrap_or(default).to_string()
     }
 
     /// Parse the network settings from the environments.toml file and set `STELLAR_RPC_URL` and
@@ -192,7 +199,7 @@ impl Args {
                 .await?;
 
                 eprintln!("🍽️ importing {:?} contract", name.clone());
-                let allow_http = if self.loam_env() == "development" {
+                let allow_http = if self.loam_env(LoamEnv::Production) == "development" {
                     "\n  allowHttp: true,"
                 } else {
                     ""
--- a/crates/loam-cli/src/commands/build/mod.rs
+++ b/crates/loam-cli/src/commands/build/mod.rs
@@ -136,7 +136,7 @@ impl Cmd {
                 cmd.arg("--release");
             } else if profile == "debug" {
             } else {
-                cmd.arg(format!("--profile={}", profile));
+                cmd.arg(format!("--profile={profile}"));
             }
             if self.all_features {
                 cmd.arg("--all-features");
--- a/crates/loam-cli/src/commands/build/mod.rs
+++ b/crates/loam-cli/src/commands/build/mod.rs
@@ -92,11 +92,17 @@ pub enum Error {
 }
 
 impl Cmd {
-    pub async fn run(&self) -> Result<(), Error> {
-        let working_dir = env::current_dir().map_err(Error::GettingCurrentDir)?;
+    pub fn list_packages(&self) -> Result<Vec<Package>, Error> {
         let metadata = self.metadata()?;
         let packages = self.packages(&metadata)?;
         let packages = loam_build::deps::get_workspace(&packages)?;
+        Ok(packages)
+    }
+
+    pub async fn run(&self) -> Result<(), Error> {
+        let working_dir = env::current_dir().map_err(Error::GettingCurrentDir)?;
+        let metadata = self.metadata()?;
+        let packages = self.list_packages()?;
         if self.list {
             for p in packages {
                 println!("{}", p.name);
--- a/crates/loam-cli/src/commands/build/mod.rs
+++ b/crates/loam-cli/src/commands/build/mod.rs
@@ -38,8 +38,8 @@ pub struct Cmd {
     #[arg(long)]
     pub package: Option<String>,
     /// Build with the specified profile
-    #[arg(long, default_value = "release")]
-    pub profile: String,
+    #[arg(long)]
+    pub profile: Option<String>,
     /// Build with the list of features activated, space or comma separated
     #[arg(long, help_heading = "Features")]
     pub features: Option<String>,
@@ -131,10 +131,12 @@ impl Cmd {
             ));
             cmd.arg("--crate-type=cdylib");
             cmd.arg("--target=wasm32-unknown-unknown");
-            if self.profile == "release" {
+            let profile = self.profile.clone().unwrap_or("release".to_string());
+            if profile == "release" {
                 cmd.arg("--release");
+            } else if profile == "debug" {
             } else {
-                cmd.arg(format!("--profile={}", self.profile));
+                cmd.arg(format!("--profile={}", profile));
             }
             if self.all_features {
                 cmd.arg("--all-features");
@@ -173,7 +175,7 @@ impl Cmd {
                 let file = format!("{}.wasm", p.name.replace('-', "_"));
                 let target_file_path = Path::new(target_dir)
                     .join("wasm32-unknown-unknown")
-                    .join(&self.profile)
+                    .join(profile)
                     .join(&file);
                 let out_file_path = out_dir.join(&file);
                 if !out_file_path.exists() {
--- a/crates/loam-cli/src/commands/build/mod.rs
+++ b/crates/loam-cli/src/commands/build/mod.rs
@@ -134,8 +134,7 @@ impl Cmd {
             let profile = self.profile.clone().unwrap_or("release".to_string());
             if profile == "release" {
                 cmd.arg("--release");
-            } else if profile == "debug" {
-            } else {
+            } else if profile != "debug" {
                 cmd.arg(format!("--profile={profile}"));
             }
             if self.all_features {
--- a/crates/loam-cli/src/commands/build/mod.rs
+++ b/crates/loam-cli/src/commands/build/mod.rs
@@ -95,8 +95,7 @@ impl Cmd {
     pub fn list_packages(&self) -> Result<Vec<Package>, Error> {
         let metadata = self.metadata()?;
         let packages = self.packages(&metadata)?;
-        let packages = loam_build::deps::get_workspace(&packages)?;
-        Ok(packages)
+        Ok(loam_build::deps::get_workspace(&packages)?)
     }
 
     pub async fn run(&self) -> Result<(), Error> {
--- a/crates/loam-cli/src/commands/dev/mod.rs
+++ b/crates/loam-cli/src/commands/dev/mod.rs
@@ -1,12 +1,13 @@
 use clap::Parser;
 use notify::{self, RecursiveMode, Watcher};
 use std::{
-    env, fs, path::{Path, PathBuf}, sync::Arc
+    env, fs,
+    path::{Path, PathBuf},
+    sync::Arc,
 };
 use tokio::sync::mpsc;
-use tokio::time;
 use tokio::sync::Mutex;
-
+use tokio::time;
 
 use crate::commands::build;
 
@@ -51,7 +52,7 @@ fn is_parent_in_watched_dirs(parent: &Path, watched_dirs: &[Arc<PathBuf>]) -> bo
 
 fn is_temporary_file(path: &Path) -> bool {
     let file_name = path.file_name().and_then(|s| s.to_str()).unwrap_or("");
-    
+
     // Vim temporary files
     if file_name.starts_with(".") {
         return true;
@@ -134,7 +135,10 @@ impl Cmd {
                                 &watched_dirs_clone,
                             );
                             // skip if the file is in the parent directory of environments.toml and it is not environments.toml
-                            if !(parent_is_env_toml_parent && !path_is_env_toml && !parent_is_in_watched_dirs) {
+                            if !(parent_is_env_toml_parent
+                                && !path_is_env_toml
+                                && !parent_is_in_watched_dirs)
+                            {
                                 println!("File changed: {path:?}");
                                 if let Err(e) = tx.blocking_send(Message::FileChanged) {
                                     eprintln!("Error sending through channel: {e}");
@@ -178,12 +182,12 @@ impl Cmd {
     async fn debounced_rebuild(self, rebuild_state: Arc<Mutex<RebuildState>>) {
         // Debounce to avoid multiple rapid rebuilds
         time::sleep(std::time::Duration::from_secs(1)).await;
-    
+
         println!("Changes detected. Rebuilding...");
         if let Err(e) = self.build().await {
             eprintln!("Build error: {e}");
         }
-    
+
         let mut state = rebuild_state.lock().await;
         state.pending = false;
     }
--- a/crates/loam-cli/src/commands/dev/mod.rs
+++ b/crates/loam-cli/src/commands/dev/mod.rs
@@ -37,6 +37,31 @@ fn is_parent_in_watched_dirs(parent: &Path, watched_dirs: &[Arc<PathBuf>]) -> bo
     watched_dirs.iter().any(|p| canonicalize_path(p) == parent)
 }
 
+fn is_temporary_file(path: &Path) -> bool {
+    let file_name = path.file_name().and_then(|s| s.to_str()).unwrap_or("");
+    
+    // Vim temporary files
+    if file_name.starts_with(".") {
+        return true;
+    }
+    if file_name.ends_with("~") {
+        return true;
+    }
+
+    // Emacs temporary files
+    if file_name.starts_with("#") && file_name.ends_with("#") {
+        return true;
+    }
+    // VSCode temporary files
+    if file_name.ends_with(".tmp") {
+        return true;
+    }
+
+    // Add more patterns for other editors as needed
+
+    false
+}
+
 impl Cmd {
     pub async fn run(&self) -> Result<(), Error> {
         let (tx, mut rx) = mpsc::channel(100);
@@ -82,9 +107,12 @@ impl Cmd {
                             | notify::EventKind::Remove(_)
                     ) {
                         if let Some(path) = event.paths.first() {
+                            // Ignore temporary files
+                            if is_temporary_file(path) {
+                                return;
+                            }
                             let env_toml_parent_abs = canonicalize_path(&env_toml_parent_clone);
                             let env_toml_path_abs = canonicalize_path(&env_toml_path_clone);
-
                             let parent_is_env_toml_parent =
                                 path.parent() == Some(env_toml_parent_abs.as_path());
                             let path_is_env_toml = path == env_toml_path_abs.as_path();
--- a/crates/loam-cli/src/commands/dev/mod.rs
+++ b/crates/loam-cli/src/commands/dev/mod.rs
@@ -29,14 +29,13 @@ impl Cmd {
         // Set LOAM_ENV to development
         std::env::set_var("LOAM_ENV", "development");
         let (tx, mut rx) = tokio::sync::mpsc::channel(100);
-        let tx_clone = tx.clone();
         let mut watcher =
             notify::recommended_watcher(move |res: Result<notify::Event, notify::Error>| {
                 if let Ok(event @ notify::Event { kind: notify::EventKind::Modify(_), .. }) = res {
                     if let Some(path) = event.paths.first() {
                         eprintln!("File modified: {:?}", path);
                         // Send a signal through the channel to trigger a rebuild
-                        if let Err(e) = tx_clone.blocking_send(()) {
+                        if let Err(e) = tx.blocking_send(()) {
                             eprintln!("Error sending through channel: {}", e);
                         }
                     }
@@ -93,7 +92,7 @@ impl Cmd {
             .build_clients
             .env
             .get_or_insert(LoamEnv::Development);
-        build_cmd.profile = "debug".to_string();
+        build_cmd.profile = "dev".to_string();
         build_cmd.run().await?;
         Ok(())
     }
--- a/crates/loam-cli/src/commands/dev/mod.rs
+++ b/crates/loam-cli/src/commands/dev/mod.rs
@@ -52,25 +52,29 @@ fn is_parent_in_watched_dirs(parent: &Path, watched_dirs: &[Arc<PathBuf>]) -> bo
 }
 
 fn is_temporary_file(path: &Path) -> bool {
+    const IGNORED_EXTENSIONS: &[&str] = &["tmp", "swp", "swo"];
     let file_name = path.file_name().and_then(|s| s.to_str()).unwrap_or("");
 
-    // Vim temporary files
-    if file_name.starts_with('.') {
+    // Vim and vscode temporary files
+    if std::path::Path::new(file_name)
+        .extension()
+        .and_then(|ext| ext.to_str())
+        .map_or(false, |ext| {
+            IGNORED_EXTENSIONS
+                .iter()
+                .any(|&ignored| ext.eq_ignore_ascii_case(ignored))
+        })
+    {
         return true;
     }
+
+    // Vim temporary files
     if file_name.ends_with('~') {
         return true;
     }
 
     // Emacs temporary files
     if file_name.starts_with('#') && file_name.ends_with('#') {
-        return true;
-    }
-    // VSCode temporary files
-    if std::path::Path::new(file_name)
-        .extension()
-        .map_or(false, |ext| ext.eq_ignore_ascii_case("tmp"))
-    {
         return true;
     }
 
--- a/crates/loam-cli/src/commands/dev/mod.rs
+++ b/crates/loam-cli/src/commands/dev/mod.rs
@@ -71,7 +71,7 @@ fn is_temporary_file(path: &Path) -> bool {
 }
 
 impl Cmd {
-    pub async fn run(&mut self) -> Result<(), Error> {
+    pub async fn run(&self) -> Result<(), Error> {
         let (tx, mut rx) = mpsc::channel::<String>(100);
         let rebuild_state = Arc::new(Mutex::new(false));
         let workspace_root: &Path = self
@@ -150,26 +150,19 @@ impl Cmd {
             watcher.watch(package_path.as_path(), RecursiveMode::Recursive)?;
         }
 
-        let build_command = self.cloned_build_command()?;
-        let cmd = build_command.lock().await;
-        if let Err(e) = cmd.run().await {
+        if let Err(e) = self.build().await {
             eprintln!("Build error: {e}");
         }
         println!("Watching for changes. Press Ctrl+C to stop.");
 
-        let rebuild_state_clone = rebuild_state.clone();
+        let rebuild_state_clone = Arc::clone(&rebuild_state);
         loop {
             tokio::select! {
                 _ = rx.recv() => {
                     let mut state = rebuild_state_clone.lock().await;
                     if !*state {
                         *state= true;
-                        // Capture the necessary clones to move into the async block
-                        let rebuild_state_clone_inner = rebuild_state_clone.clone();
-                        let inner_build_command = self.cloned_build_command()?;
-                        tokio::spawn(async move {
-                            Self::debounced_rebuild(inner_build_command, rebuild_state_clone_inner).await;
-                        });
+                        tokio::spawn(self.clone().debounced_rebuild(Arc::clone(&rebuild_state_clone)));
                     }
                 }
                 _ = tokio::signal::ctrl_c() => {
@@ -181,13 +174,12 @@ impl Cmd {
         Ok(())
     }
 
-    async fn debounced_rebuild(build_cmd: Arc<Mutex<build::Cmd>>, rebuild_state: Arc<Mutex<bool>>) {
+    async fn debounced_rebuild(self, rebuild_state: Arc<Mutex<bool>>) {
         // Debounce to avoid multiple rapid rebuilds
         time::sleep(std::time::Duration::from_secs(1)).await;
 
         println!("Changes detected. Rebuilding...");
-        let cmd = build_cmd.lock().await;
-        if let Err(e) = cmd.run().await {
+        if let Err(e) = self.build().await {
             eprintln!("Build error: {e}");
         }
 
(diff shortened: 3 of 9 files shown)


crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+                }
+            })
+            .unwrap();
+
+        let workspace_root = self
+            .build_cmd
+            .manifest_path
+            .parent()
+            .map(|path| path.to_path_buf())
+            .unwrap_or_else(|| PathBuf::from("."));
```
willemneal, line 54: You should be able to do `Path::new(".")` and avoid needing the map. `workspace_root` can be a `Path`

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        let env_toml = workspace_root.join("environments.toml");
+        if env_toml.exists() {
```
willemneal, line 56: Does watching a non-existent file cause an error? Otherwise it would seem like you would want to watch for the files creation as an update.

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        // Set LOAM_ENV to development
+        std::env::set_var("LOAM_ENV", "development");
+        let (tx, mut rx) = tokio::sync::mpsc::channel(100);
+        let tx_clone = tx.clone();
+        let mut watcher =
+            notify::recommended_watcher(move |res: Result<notify::Event, notify::Error>| {
+                if let Ok(event) = res {
+                    if let notify::EventKind::Modify(_) = event.kind {
+                        if let Some(path) = event.paths.first() {
+                            println!("File modified: {:?}", path);
```
willemneal, line 38:
```suggestion
                            eprintln!("File modified: {:?}", path);
```

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+
+impl Cmd {
+    pub async fn run(&self) -> Result<(), Error> {
+        // Set LOAM_ENV to development
+        std::env::set_var("LOAM_ENV", "development");
+        let (tx, mut rx) = tokio::sync::mpsc::channel(100);
+        let tx_clone = tx.clone();
+        let mut watcher =
+            notify::recommended_watcher(move |res: Result<notify::Event, notify::Error>| {
+                if let Ok(event) = res {
```
willemneal, line 35:
```suggestion
                if let Ok(event @ notify::Event { kind: notify::EventKind::Modify(_) } ) = res {
```
Since the assignment is a pattern match you can match on the inner type and the `@` lets you simultaneously assign the value to a variable. So this has the same effect as the nested ifs. See example here:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a2ae21a539c81d4d157644d6da6b978b

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    #[error("IO error: {0}")]
+    Io(#[from] std::io::Error),
+}
+
+impl Cmd {
+    pub async fn run(&self) -> Result<(), Error> {
+        // Set LOAM_ENV to development
+        std::env::set_var("LOAM_ENV", "development");
+        let (tx, mut rx) = tokio::sync::mpsc::channel(100);
+        let tx_clone = tx.clone();
```
willemneal, line 32: What errors were you getting if this wasn't cloned?

crates/loam-cli/src/commands/mod.rs
```diff
@@ -39,6 +40,7 @@ impl Root {
         match &mut self.cmd {
```
willemneal, line 40: Ah so since this a mutable reference you can change `dev_info.run` to use `(&mut self)`, which removes the need to clone in the `build` method.

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+        Ok(())
+    }
+
+    async fn build(&self) -> Result<(), Error> {
+        let mut build_cmd = self.build_cmd.clone();
+        build_cmd
+            .build_clients
+            .env
+            .get_or_insert(LoamEnv::Development);
+        build_cmd.profile = "dev".to_string();
```
willemneal, line 99: @chadoh Do you ever expect someone to use a non-dev profile? Currently the default is set by clap, but we could change to an Option like `env` which would let us respect what the user pasted in and not overwrite it here.

willemneal, line 99:
```suggestion
        build_cmd.profile = "debug".to_string();
```

crates/loam-cli/src/commands/dev/mod.rs
willemneal: Also this suggestion is not good on it's own since it doesn't remove the inner if.

willemneal: Also since these types don't collide with any in this file I'd just import them for better readability.

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: yes, `error: No such file or directory`

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: got it

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: done

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: It works if it isn't cloned. There would be  a problem if tx was used later in the code but since it isn't I don't think we need to clone it.

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: Seems to not like using the name debug: 
```
error: profile name `debug` is reserved
To configure the default development profile, use the name `dev` as in [profile.dev]
```

willemneal: Ah yeah. `cargo build` is debug by default, you need to add either `--release` or a `--profile <defined in Cargo.toml>`. So this is more reason to make this an option and then it will be debug by default.

BlaineHeffron: I see. I changed it to be an option and default to debug. If its debug it doesnt add the --profile command.

crates/loam-cli/src/commands/build/mod.rs
```diff
@@ -125,10 +131,12 @@ impl Cmd {
...
             cmd.arg("--target=wasm32-unknown-unknown");
-            if self.profile == "release" {
+            let profile = self.profile.clone().unwrap_or("release".to_string());
+            if profile == "release" {
                 cmd.arg("--release");
+            } else if profile == "debug" {
             } else {
-                cmd.arg(format!("--profile={}", self.profile));
+                cmd.arg(format!("--profile={profile}"));
             }
```
chadoh, line 140: if it is neither `release` nor `debug` then we ignore it?

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[error(transparent)]
+    Build(#[from] build::Error),
+    #[error("IO error: {0}")]
+    Io(#[from] std::io::Error),
+}
+
+impl Cmd {
+    pub async fn run(&self) -> Result<(), Error> {
+        // Set LOAM_ENV to development
+        std::env::set_var("LOAM_ENV", "development");
```
chadoh, line 30: I think we should _default_ `LOAM_ENV` to `development` _when you use `loam dev`_. And when you use `loam build`, we should default it to `production`. But you should be able to override the defaults.

I think this is the behavior of other commands like `astro dev` and `astro build`.

If you want to run the watcher but compile everything for `LOAM_ENV=production`, why should we stop you?

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+            })
+            .unwrap();
+
+        let workspace_root: &Path = self
+            .build_cmd
+            .manifest_path
+            .parent()
+            .unwrap_or_else(|| Path::new("."));
+        let env_toml = workspace_root.join("environments.toml");
+        if env_toml.exists() {
```
chadoh, line 58: In addition to noticing if you _modify_ `environments.toml`, can we also check if you _add_ or _delete_ it?

chadoh (changes requested):
Some questions/comments. Also, I want to see some tests!

crates/loam-cli/src/commands/build/mod.rs
willemneal: No we ignore if it is debug. Or said another way don't ignore it if isn't debug.

crates/loam-cli/src/commands/build/mod.rs
```diff
@@ -125,10 +131,12 @@ impl Cmd {
             ));
             cmd.arg("--crate-type=cdylib");
             cmd.arg("--target=wasm32-unknown-unknown");
-            if self.profile == "release" {
+            let profile = self.profile.clone().unwrap_or("release".to_string());
+            if profile == "release" {
                 cmd.arg("--release");
+            } else if profile == "debug" {
             } else {
```
willemneal, line 138:
```suggestion
            } else if profile != "debug" {
```

crates/loam-cli/src/commands/build/mod.rs
```diff
@@ -125,10 +131,12 @@ impl Cmd {
             ));
             cmd.arg("--crate-type=cdylib");
             cmd.arg("--target=wasm32-unknown-unknown");
-            if self.profile == "release" {
+            let profile = self.profile.clone().unwrap_or("release".to_string());
```
willemneal, line 134: You can use `.as_ref` on an option to go from `Option<T>` to `Option<&T>` and in this case you can use `as_deref` would would go `Option<String>` to `Option<&str>` preventing the need to clone and use a string.
Second you should have used `unwrap_or_else` because this way you call `release.to_string` even if the `Option` is a `Some`.

```suggestion
            let profile = self.profile.as_deref().unwrap_or("release");
```

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: Good catch, I should have deleted this line. It is currently defaulting to development later when it runs build.

BlaineHeffron:
Setting to draft while I working on watching potential additions / deletions of environments.toml and adding unit tests.

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: I modified the code to do this. It was a pretty substantial change because often `environments.toml` is in a  directory not otherwise watched. To do this I had to watch its parent directory but then make sure only changes to that file were made (assuming its not in a directory already being watched for changes). I used Arc<PathBuf> and Arc::clone to do this because the lifetimes of the paths being passed to the watcher function were causing me issues.

BlaineHeffron:
Some substantial changes have been made. First, I found out while testing that the debounce mechanism wasn't working properly. A mutex state is used to track whether or not rebuilding should be blocked. This makes it so that if multiple files are changed while the build process is ongoing, it will only trigger one rebuild at the end. Before it would trigger N rebuilds where N were the number of changes detected.

Second, the logic for determining changes was overhauled. File modifications, deletions, and insertions in the watched directories all trigger rebuilds. Same with `environments.toml`. I needed to ensure that files in the same directory as `environments.toml` are ignored (this is because now it must watch the parent directory of `environments.toml` to determine if creations/deletions are made), so I needed to use `Arc` to pass the path objects into the watcher.

Finally, I added tests. They are all in one test block because the initial build of the directory takes some time. This modifies environments.toml and makes sure its loading in the new configuration. It also changes a file within one of the contracts directory and ensures it is detected and rebuilt.

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+
+use super::build::clients::LoamEnv;
+
+struct RebuildState {
+    pending: bool,
+}
+
+enum Message {
+    FileChanged,
+}
```
chadoh, line 22: an enum with one variant seems odd

crates/loam-cli/tests/it/build_clients/dev.rs
```diff
@@ -0,0 +1,142 @@
...
+            .await;
+            dev_process
+                .kill()
+                .await
+                .expect("Failed to kill dev process");
+
+            dev_process
+                .wait()
+                .await
+                .expect("Failed to wait for dev process");
```
chadoh, line 112: What's happening here?

chadoh (approved):
Wow so glad I asked for tests! Two small questions below. I also wonder if it's worth testing the "environments.toml added/removed" cases.

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+use tokio::sync::mpsc;
+use tokio::sync::Mutex;
+use tokio::time;
+
+use crate::commands::build;
+
+use super::build::clients::LoamEnv;
+
+struct RebuildState {
+    pending: bool,
```
willemneal, line 17: Similar question as chad here. Is this needed or can it just be a bool?

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+        for package_path in watched_dirs {
+            watcher.watch(package_path.as_path(), RecursiveMode::Recursive)?;
+        }
+
+        if let Err(e) = self.build().await {
+            eprintln!("Build error: {e}");
+        }
+        println!("Watching for changes. Press Ctrl+C to stop.");
+
+        let rebuild_state_clone = Arc::clone(&rebuild_state);
```
willemneal, line 166:
```suggestion
        let rebuild_state_clone = rebulid_state.clone();
```
Does this not work?

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+        return true;
+    }
+
+    // Add more patterns for other editors as needed
+
+    false
+}
+
+impl Cmd {
+    pub async fn run(&self) -> Result<(), Error> {
```
willemneal, line 82:
```suggestion
    pub async fn run(&mut self) -> Result<(), Error> {
```

willemneal: We can do this because the caller has a mutable reference! Then you can do.
```
         self.build_cmd
            .build_clients
            .env
            .get_or_insert(LoamEnv::Development);
        self.build_cmd.profile.get_or_insert_with(|| "debug".to_string());
        let build_cmd = Arc::new(build_cmd.clone());
```

Then below you can replace `self.clone()` with `build_cmd.clone()`, which just increments the counter and creates a new reference.

crates/loam-cli/src/commands/dev/mod.rs
willemneal: Also do you need this at all? Can't you just `rebuild_state.clone()` below each time?

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+            }
+        }
+        Ok(())
+    }
+
+    async fn debounced_rebuild(self, rebuild_state: Arc<Mutex<RebuildState>>) {
+        // Debounce to avoid multiple rapid rebuilds
+        time::sleep(std::time::Duration::from_secs(1)).await;
+
+        println!("Changes detected. Rebuilding...");
```
willemneal, line 189:
```suggestion
        eprintln!("Changes detected. Rebuilding...");
```

crates/loam-cli/tests/it/build_clients/dev.rs
```diff
@@ -0,0 +1,142 @@
...
+            sleep(Duration::from_millis(100)).await;
+        }
+    })
+    .await;
+    match result {
+        Ok(_) => {
+            println!("Found string {}", expected)
+        }
+        Err(_) => panic!("Timed out waiting for output: {}", expected),
+    }
```
willemneal, line 141:
```suggestion
    match result {
        Ok(_) => {
            println!("Found string {expected}")
        }
        Err(_) => panic!("Timed out waiting for output: {expected}"),
    }
```

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: fixed

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: changed to just a bool

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: Okay, refactored to only clone build_cmd. Let me know what you think.

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: should all the println statements be eprintln?

crates/loam-cli/tests/it/build_clients/dev.rs
BlaineHeffron: It doesn't work, doesnt this require a string format call or something?

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: I reverted the changes because it was causing a failure during the build command. I'm not sure why what I did didn't work. If you think its important I can make another attempt at it.

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -143,7 +135,7 @@ impl Cmd {
                                 && !parent_is_in_watched_dirs)
                             {
                                 println!("File changed: {path:?}");
-                                if let Err(e) = tx.blocking_send(Message::FileChanged) {
+                                if let Err(e) = tx.blocking_send("FileChanged".to_string()) {
```
willemneal, line 138: I actually prefer the enum with a single case over a string. For one it's less to send.

crates/loam-cli/tests/it/build_clients/dev.rs
BlaineHeffron: oops, good eye. I removed the unnecessary call at the end.

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: Yes, I changed it, thanks.

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: Okay, I'll go with it.

BlaineHeffron:
It has been refactored to only clone the build command. I also fixed an issue where environments.toml deletions were not being detected. I added a test case for that. All comments have been addressed.

chadoh (approved):
Heroic

chadoh:
Go ahead and get the PR title and description, or maybe the squash commit message (you'll have to check the set up) into a format that can be used by the auto release logic. And then merge away!

crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -62,8 +69,8 @@ impl Args {
         Ok(())
     }
 
-    fn loam_env(self) -> String {
-        format!("{:?}", self.env).to_lowercase()
+    fn loam_env(self, default: LoamEnv) -> String {
+        self.env.unwrap_or(default).to_string()
```
willemneal, line 73:
```suggestion
        self.env.unwrap_or(default).to_lowercase()
```
It returns a `String` and we still want to make sure it's lowercase.

crates/loam-cli/src/commands/build/mod.rs
```diff
@@ -92,11 +92,17 @@ pub enum Error {
 }
 
 impl Cmd {
-    pub async fn run(&self) -> Result<(), Error> {
-        let working_dir = env::current_dir().map_err(Error::GettingCurrentDir)?;
+    pub fn list_packages(&self) -> Result<Vec<Package>, Error> {
         let metadata = self.metadata()?;
         let packages = self.packages(&metadata)?;
         let packages = loam_build::deps::get_workspace(&packages)?;
+        Ok(packages)
```
willemneal, line 99:
```suggestion
        Ok(loam_build::deps::get_workspace(&packages)?)
```

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+                            | notify::EventKind::Modify(_)
+                            | notify::EventKind::Remove(_)
+                    ) {
+                        if let Some(path) = event.paths.first() {
+                            // Ignore temporary files
+                            if is_temporary_file(path) {
+                                return;
+                            }
+                            let env_toml_parent_abs = canonicalize_path(&env_toml_parent_clone);
+                            let env_toml_path_abs = canonicalize_path(&env_toml_path_clone);
```
willemneal, line 133: Can this be done outside?

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+
+fn is_parent_in_watched_dirs(parent: &Path, watched_dirs: &[Arc<PathBuf>]) -> bool {
+    watched_dirs.iter().any(|p| canonicalize_path(p) == parent)
+}
+
+fn is_temporary_file(path: &Path) -> bool {
+    let file_name = path.file_name().and_then(|s| s.to_str()).unwrap_or("");
+
+    // Vim temporary files
+    if file_name.starts_with('.') {
```
willemneal, line 58: This could also be an `.env` file. Do they end with `.swp`?

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+    if file_name.ends_with('~') {
+        return true;
+    }
+
+    // Emacs temporary files
+    if file_name.starts_with('#') && file_name.ends_with('#') {
+        return true;
+    }
+    // VSCode temporary files
+    if std::path::Path::new(file_name)
```
willemneal, line 70:
```suggestion
    if Path::new(file_name)
```

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+                                return;
+                            }
+                            let env_toml_parent_abs = canonicalize_path(&env_toml_parent_clone);
+                            let env_toml_path_abs = canonicalize_path(&env_toml_path_clone);
+                            let parent_is_env_toml_parent =
+                                path.parent() == Some(env_toml_parent_abs.as_path());
+                            let path_is_env_toml = path == env_toml_path_abs.as_path();
+                            let parent_is_in_watched_dirs = is_parent_in_watched_dirs(
+                                &env_toml_parent_abs,
+                                &watched_dirs_clone,
```
willemneal, line 139: This seems like it could also be computed outside the closure.  I'm I don't understand why this is an issue.

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+                            );
+                            // skip if the file is in the parent directory of environments.toml and it is not environments.toml
+                            if !(parent_is_env_toml_parent
+                                && !path_is_env_toml
+                                && !parent_is_in_watched_dirs)
```
willemneal, line 144:
```suggestion
                            if !parent_is_env_toml_parent
                                || path_is_env_toml
                                || parent_is_in_watched_dirs
```
It's best to distribute the negation as it's less work for the reader.
So in this case if `parent_is_in_watched_dirs` is true which is independent of the `path` argument this is invoked?

willemneal:
I have one thought that would help clean this up a bit.

```rust
#[derive(Clone)
struct Watcher {
    root_env: Arc<PathBuf>,
    packages: Arc<Vec<PathBuf>,
}

impl Watcher {
    pub fn new(root_env: &Path, packages: &[PathBuf]) -> Self {
        Self {
            root_env: Arc::new(conanicalize_path(root_env))
            packages: Arc::new(packages.iter().map(canonicalize_path).collect())
        }
    }
/// Relevant methods here, e.g.
    pub fn is_env_toml_parent(path: &Path) -> bool {
        self.root_env.parent() == path.parent()
    }
}

```

Since `Arc` implements `Clone` we can derive it for this struct. Then it can be `.clone()` and the relevant logic placed inside.

crates/loam-cli/src/commands/build/clients.rs
BlaineHeffron: good catch

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: `.swp` wasn't the only one but I think `.swo`, `~` and `.swp` should cover it all.

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: Fixed

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: yes because any file in the watched dirs should trigger a rebuild.

BlaineHeffron: basically we have to add the parent dir of the env file to the list of dirs in order to check if there is a deletion or creation of the env file. If that dir is already in the watched dirs then all watched files should trigger a rebuild

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+struct Watcher {
+    root_env: Arc<PathBuf>,
+    packages: Arc<Vec<PathBuf>>,
+}
+
+impl Watcher {
+    pub fn new(root_env: &Path, packages: &[PathBuf]) -> Self {
+        Self {
+            root_env: Arc::new(canonicalize_path(root_env)),
+            packages: Arc::new(packages.iter().map(|p| canonicalize_path(p)).collect()),
```
willemneal, line 60:
```suggestion
            packages: Arc::new(packages.iter().map(canonicalize_path).collect()),
```

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    pub fn is_env_toml(&self, path: &Path) -> bool {
+        canonicalize_path(path) == *self.root_env
+    }
+}
+
+fn is_temporary_file(path: &Path) -> bool {
+    const IGNORED_EXTENSIONS: &[&str] = &["tmp", "swp", "swo"];
+    let file_name = path.file_name().and_then(|s| s.to_str()).unwrap_or("");
```
willemneal, line 76: The default af `""` seems strange. The reason you need to convert to a string is that a `str` is UTF-8 whereas  a Path uses an OsStr because some OS's can have path names which aren't safely converted to UTF-8. So it's trying to convert to a string and returning Option of whether it's valid UTF-8. We could consider using [camino](https://docs.rs/camino/latest/camino) which is a crate that enforces paths be UTF-8 (which for most rust projects is definitely the case). I've been wanting to try it out because then you can go between a Path and `str` without any headaches.

But for now I'd use an `expect` and we could use the other crate in a follow up PR.  I think it would clean up the whole repo.

willemneal, line 76:
```suggestion
    let file_name = path.file_name().and_then(Path::to_str).unwrap_or("");
```

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+            .map(|package| PathBuf::from(package.manifest_path.parent().unwrap().as_str()))
+            .collect::<Vec<_>>();
+
+        let watcher = Watcher::new(&env_toml_path, &packages);
+
+        for package_path in watcher.packages.iter() {
+            eprintln!("Watching {}", package_path.display());
+        }
+
+        let watcher_clone = watcher.clone();
```
willemneal, line 130: Why still the initial clone?

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: I realized the parent_is_env_toml_parent check is redundant so its been removed in the refactored version

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+                        event.kind,
+                        notify::EventKind::Create(_)
+                            | notify::EventKind::Modify(_)
+                            | notify::EventKind::Remove(_)
+                    ) {
+                        if let Some(path) = event.paths.first() {
+                            if is_temporary_file(path) {
+                                return;
+                            }
+                            if watcher_clone.is_watched(path) || watcher_clone.is_env_toml(path) {
```
willemneal, line 144:
```suggestion
                            let watcher = watcher.clone();
                            if watcher.is_watched(path) || watcher.is_env_toml(path) {
```

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+            .manifest_path
+            .parent()
+            .unwrap_or_else(|| Path::new("."));
+        let env_toml_path = workspace_root.join("environments.toml");
+
+        let packages = self
+            .build_cmd
+            .list_packages()?
+            .into_iter()
+            .map(|package| PathBuf::from(package.manifest_path.parent().unwrap().as_str()))
```
willemneal, line 121: Isn't `parent` an `Option<&Path>`?
```suggestion
            .map(|package| package.manifest_path.parent().unwrap().into_path_buf())
```

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+            .collect::<Vec<_>>();
+
+        let watcher = Watcher::new(&env_toml_path, &packages);
+
+        for package_path in watcher.packages.iter() {
```
willemneal, line 126:
```suggestion
        for package_path in &watcher.packages {
```

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+                    }
+                }
+            })
+            .unwrap();
+
+        notify_watcher.watch(
+            watcher.root_env.parent().unwrap(),
+            RecursiveMode::NonRecursive,
+        )?;
+        for package_path in watcher.packages.iter() {
```
willemneal, line 160:
```suggestion
        for package_path in &watcher.packages {
```

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+            notify_watcher.watch(package_path, RecursiveMode::Recursive)?;
```
willemneal, line 161: It would be nice to be able to filter out the temporary files here. But that is an optimization we can deal with later (though I doubt we'll ever need)

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: Seems to complain about the type: ``` note: expected function signature `fn(&PathBuf) -> _`
               found function signature `fn(&std::path::Path) -> _`
note: required by a bound in `std::iter::Iterator::map`
   --> /home/blaine/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:764:12
    |
761 |     fn map<B, F>(self, f: F) -> Map<Self, F>
    |        --- required by a bound in this associated function
...
764 |         F: FnMut(Self::Item) -> B,
    |            ^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Iterator::map`
help: consider wrapping the function in a closure```

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: Okay, set it to use `expect` for now.

BlaineHeffron: `file_name` returns an `Option<&OsStr>` which isn't compatible with `Path::to_str`

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: ```error[E0382]: borrow of moved value: `watcher`
    --> crates/loam-cli/src/commands/dev/mod.rs:163:29
     |
127  |         let watcher = Watcher::new(&env_toml_path, &packages);
     |             ------- move occurs because `watcher` has type `dev::Watcher`, which does not implement the `Copy` trait
...
135  |             notify::recommended_watcher(move |res: Result<notify::Event, notify::Error>| {
     |                                         ------------------------------------------------ value moved into closure here
...
147  |                             if watcher.is_watched(path) || watcher.is_env_toml(path) {
     |                                ------- variable moved due to use in closure
...
163  |         for package_path in watcher.packages.iter() {
     |                             ^^^^^^^^^^^^^^^^^^^^^^^ value borrowed here after move
     |
     = note: borrow occurs due to deref coercion to `Vec<PathBuf>`
note: deref defined here
    --> /home/blaine/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/sync.rs:2102:5
     |
2102 |     type Target = T;
     |     ^^^^^^^^^^^
```

Comment (willemneal) [reply crates/loam-cli/src/commands/dev/mod.rs:133]: Yes but why not clone it inside the closure as I have below?

Comment (BlaineHeffron) [reply crates/loam-cli/src/commands/dev/mod.rs:133]: Looks like I can avoid cloning if I use the package and env paths directly when setting the notify_watcher to watch them.

File: crates/loam-cli/src/commands/dev/mod.rs (hunk 32 above)
Comment (BlaineHeffron) [reply crates/loam-cli/src/commands/dev/mod.rs:130]: Seems the `to_path_buf` method returns a `Utf8PathBuf` which we'd then have to refactor alot of the code to pass that type. Is that what you'd prefer?

File: crates/loam-cli/src/commands/dev/mod.rs (hunk 33 above)
Comment (BlaineHeffron) [reply crates/loam-cli/src/commands/dev/mod.rs:135]: This results in the following error: 
```
&Arc<Vec<PathBuf>>` is not an iterator
the trait `Iterator` is not implemented for `&Arc<Vec<PathBuf>>`, which is required by `&Arc<Vec<PathBuf>>: IntoIterator`
required for `&Arc<Vec<PathBuf>>` to implement `IntoIterator
```

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: Same issue as above

crates/loam-cli/src/commands/dev/mod.rs
BlaineHeffron: I think the difficulty is that its watching the whole directory so that it catches new files, so we couldn't calculate if a new file matches the temporary file pattern a-priori. Maybe there is a method I'm not aware of that does that you propose.

willemneal: Yeah no worries. Again it's filtering later works just as well.

crates/loam-cli/src/commands/dev/mod.rs
willemneal: This is fine for now. Again, we can return with `camino` later to clean up all path stuff.

willemneal (approved):
Great work! A deep dive into Rust!

PR #88, opened 2024-06-24
feat(CLI): use aliases to avoid needlessly redeploying contracts

BlaineHeffron:
Resolves issue https://github.com/loambuild/loam-sdk/issues/73#issue-2358198453


crates/loam-cli/src/commands/build/build_clients.rs
```diff
@@ -123,79 +125,57 @@ impl Args {
...
-            }
-        } else {
-            Ok(None)
-        }
-    }
 
-    async fn contract_hash_matches(&self, contract_id: &str, hash: &str, network: &Network) -> Result<bool, Error> {
-        let network_args = cli::network::Args {
+
+    fn get_network_args(network: &Network) -> cli::network::Args {
```
willemneal, line 130: I think we should instead remove the local `Network` type, which should have never existed since this type exists.

BlaineHeffron: Okay, should I perform the refactor now or in a separate PR?

willemneal: I would go ahead and do it now since you need the original type.

BlaineHeffron: I have the refactor done but the deserialize is not implemented for cli::network::Args which is needed in `Environment`:

```
#[derive(Debug, serde::Deserialize, Clone)]
pub struct Environment {
    pub accounts: Option<Vec<Account>>,
    pub network: Network,
    pub contracts: Option<Map<Box<str>, Contract>>,
}
```
It looks like this could be worked around with a custom deserialize method for Environment, if you want to go that route.

willemneal: There are actually two Network types in the CLI. Here is the the one that is serialized: https://github.com/stellar/stellar-cli/blob/7984cd0d014d67ff3ca5fe720b615001b255d524/cmd/soroban-cli/src/commands/network/mod.rs#L170

The other is just for the CLI args. The idea is that that is used to resolve to a network with a `rpc_url` and `network_passphrase`, no options.

BlaineHeffron: Excellent, thanks

BlaineHeffron: Problem I'm seeing is that `Network` doesnt have the `network` argument like the `Args` version does. The `name` parameter in environments.toml is being mapped to `network`.

chadoh: Stop @willemneal! I already explained why we need our own `Network`! https://github.com/loambuild/loam-sdk/pull/42#pullrequestreview-2130862138

BlaineHeffron: I seem to have rediscovered this reasoning lol

crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -108,6 +126,83 @@ impl Args {
...
+            r#"import * as Client from '{name}';
+    import {{ rpcUrl }} from './util';
+    
+    export default new Client.Client({{
+      networkPassphrase: '{network}',
+      contractId: '{contract_id}',
+      rpcUrl,{allow_http}
+      publicKey: undefined,
+    }});
+    "#
```
chadoh, line 199: We want this to look right in a TS file

```suggestion
import {{ rpcUrl }} from './util';
    
export default new Client.Client({{
  networkPassphrase: '{network}',
  contractId: '{contract_id}',
  rpcUrl,{allow_http}
  publicKey: undefined,
}});
"#
```

crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -192,30 +305,11 @@ impl Args {
...
-                );
-                let path = workspace_root.join(format!("src/contracts/{name}.ts"));
-                std::fs::write(path, template).expect("could not write contract template");
+                self.write_contract_template(workspace_root, name, &contract_id)?;
             };
         }
 
         Ok(())
     }
+
```
chadoh, line 314: is this a clippy preference?

chadoh (changes requested):
Looks great! Love the refactoring. Just one (one and a half?) quick cleanup item(s).

crates/loam-cli/src/commands/build/clients.rs
BlaineHeffron: yes

crates/loam-cli/src/commands/build/clients.rs
BlaineHeffron: Ah, good catch, thanks!

chadoh (changes requested):
We need tests!

BlaineHeffron:
Added tests - adds it twice, ensure alias check works. Changes the file, ensures it replaces. Changes again with production, ensures error is thrown.

willemneal:
THis is a flaky test issue that appears in the CLI too. The solution I had was to add sleeps to the failing test. However, this is not the correct solution for sure. One possible reason is that the tests are all trying to install and deploy the same contract. So one solution is to add random/unique data to the wasm's custom section. This will make the wasm have a new hash. So as long as the data that is used in the custom section is the same a test than the hash check you are doing should still work.

We can chat about it if you want. I need to do it for the CLI so  would be good to figure out an easy way to do it. Maybe we can make a small crate and publish to crates.io. This way we can use it across both projects and it'll be a lesson in publishing crates!

BlaineHeffron:
Tried @willemneal suggestion, didn't seem to work. Very strange, makes me wonder if there is a different underlying issue here.

BlaineHeffron:
Turns out Willem's suggestion worked, problem was I modified the wrong wasm file (the one in the build directory instead of the one copied to `/target/loam`).

chadoh (approved):
Hooray!

PR #91, opened 2024-06-27
Feat/nft example

elizabethengelman:


examples/soroban/nft/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+#![no_std]
+use loam_sdk::soroban_contract;
+use loam_subcontract_core::{admin::Admin, Core};
+use loam_soroban_sdk::{Address, Bytes};
```
elizabethengelman, line 4: I needed to include `use loam_soroban_sdk::{Address, Bytes};` to make the code compile otherwise I was seeing this error:
```
error[E0412]: cannot find type `Bytes` in this scope
  --> examples/soroban/nft/src/lib.rs:28:1
  |
28 | soroban_contract!();
  | ^^^^^^^^^^^^^^^^^^^ not found in this scope
```

Not sure if there is a better way to handle this, as I didn't see other examples needing to do this 🤔

examples/soroban/nft/src/subcontract.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn transfer(&mut self, id: u32, current_owner: Address, new_owner: Address);
+
+    // Get the NFT with the given ID
+    fn get_nft(&self, id: u32) -> Option<Bytes>;
+
+    // Find the owner of the NFT with the given ID
+    fn get_owner(&self, id: u32) -> Option<Address>;
+
+    // Get the total count of NFTs
+    fn get_total_count(&self) -> u32;
```
willemneal, line 21:
```suggestion
    fn mint(&mut self, owner: Address, metadata: Bytes) -> u32;

    // Transfer an NFT with the given ID from current_owner to new_owner
    fn transfer(&mut self, id: u32, current_owner: loam_sdk::soroban_sdk::Address, new_owner: loam_sdk::soroban_sdk::Address);

    // Get the NFT with the given ID
    fn get_nft(&self, id: u32) -> Option<loam_sdk::soroban_sdk::Bytes>;

    // Find the owner of the NFT with the given ID
    fn get_owner(&self, id: u32) -> Option<loam_sdk::soroban_sdk::Address>;

    // Get the total count of NFTs
    fn get_total_count(&self) -> u32;
```

examples/soroban/nft/src/subcontract.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+#[subcontract]
+pub trait IsInitable {
+    fn nft_init(&mut self, admin: Address, name: Bytes);
```
willemneal, line 26:
```suggestion
    fn nft_init(&mut self, admin: loam_sdk::soroban_sdk::Address, name: loam_sdk::soroban_sdk::Bytes);
```

examples/soroban/nft/src/subcontract.rs
```diff
@@ -0,0 +1,27 @@
+use loam_sdk::{
+    soroban_sdk::{Address, Bytes, Lazy},
+    subcontract,
+};
```
willemneal, line 4:
```suggestion
use loam_sdk::subcontract;
```

examples/soroban/nft/src/lib.rs
willemneal: The issue here was that the methods from the trait are copy/pasted here. So you need to fully qualify the types to avoid needed to import them again.

```rust
impl SorobanContract {
    //...
    // Get the NFT with the given ID
    fn get_nft( id: u32) -> Option<loam_sdk::soroban_sdk::Bytes> {
        Contract::lazy_get().unwrap().get_nft(id)
    }
}
```

From the outside the trait doesn't have any `self` references. So this code is generated and to invoke the method bound to `Contract` via the `NonFungible` trait.

examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,97 @@
...
+    name: Bytes,
+    total_count: u32,
+    owners_to_nft_ids: Map<Address, u32>,
+    nft_ids_to_owners: Map<u32, Address>,
+    nft_ids_to_metadata: Map<u32, Bytes>,
+}
+
+impl MyNonFungibleToken {
+    #[must_use]
+    pub fn new(admin: Address, name: Bytes) -> Self {
```
willemneal, line 25:
```suggestion
    pub fn new(admin: Address, name: Bytes) -> Self {
        
```

examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+impl Default for MyNonFungibleToken {
+    fn default() -> Self {
+        MyNonFungibleToken::new(env().current_contract_address(), Bytes::new(env()))
+    }
+}
+
+impl IsInitable for MyNonFungibleToken {
+    fn nft_init(&mut self, admin: Address, name: Bytes) {
```
willemneal, line 44:
```suggestion
    fn nft_init(&self, admin: Address, name: Bytes) {
```
If it's `&mut` it does the following in the caller.
```rust
let mut s = Initable::Self::get_lazy().unwrap_or_default();
let res = s.nft_init(&mut s, admin, name);
Initable::Self::set_lazy(s);
res
```
So in this case it might be overwritten with the default. This does suggest that we have a bug in the implementation of init methods. Ideally, the value should initially be `None` so that calling it twice wouldn't overwrite the value. I would need to change the design and macro. But the more I think about it. It should be that in the init case it should be `None` and for all other methods it should be `Some` as otherwise means the contract hasn't been initialized.

willemneal: This shouldn't be an issue if we ensure that when say using smartdeploy it's initialized in the same transaction as its deployed.

examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,97 @@
...
+        self.nft_ids_to_owners.set(new_id, owner.clone());
+        self.owners_to_nft_ids.set(owner, new_id);
+        self.total_count = new_id;
+
+        new_id
+    }
+
+    // Transfer the NFT from the current owner to the new owner
+    fn transfer(&mut self, id: u32, current_owner: Address, new_owner: Address) {
+        if let Some(owner_id) = self.nft_ids_to_owners.get(id) {
```
willemneal, line 69:
```suggestion
        current_owner.require_auth();
        if let Some(owner_id) = self.nft_ids_to_owners.get(id) {
```

examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,97 @@
...
+                "You are not the owner of this NFT"
+            );
+            // remove the current owner
+            self.nft_ids_to_owners.remove(id);
+            self.owners_to_nft_ids.remove(current_owner);
+
+            // add the new owner
+            self.nft_ids_to_owners.set(id, new_owner.clone());
+            self.owners_to_nft_ids.set(new_owner, id);
+        }
```
willemneal, line 81: Should it panic if it doesn't exist?

examples/soroban/nft/src/lib.rs
elizabethengelman: Ohhh, that is helpful, thanks. I may need to think on this a bit more, but I think that makes sense 👌

examples/soroban/nft/src/nft.rs
elizabethengelman: Good call - just pushed a change

examples/soroban/nft/src/nft.rs
elizabethengelman: I think I understand  - `&mut self` here opens us up to the possibility that the internal state of the singleton can be overwritten. 

Are you saying that you want to redesign this so that users aren't able to get themselves into that situation by initializing a contract with a state of `None`. And then other methods require that the state is `Some`? Otherwise it will indicate that the contract hasn't been initialized. 

That all makes sense to me!

elizabethengelman: Also, i think that we have a `&mut self` in the fungible token example as well - does it make sense to update that too?

elizabethengelman: updated the fungible token example here: 80eba5b6adb9f7984eaeda4ddb51691909d0c088

examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,99 @@
...
```
willemneal, line 83:
```suggestion
        let owner_id = self.nft_ids_to_owners.get(id).expect("NFT does not exist");
        assert!(
            owner_id == current_owner,
            "You are not the owner of this NFT"
        );
        // remove the current owner
        self.nft_ids_to_owners.remove(id);
        self.owners_to_nft_ids.remove(current_owner);

        // add the new owner
        self.nft_ids_to_owners.set(id, new_owner.clone());
        self.owners_to_nft_ids.set(new_owner, id);
```

elizabethengelman: Much better!

examples/soroban/nft/src/nft.rs
```diff
@@ -15,7 +15,7 @@ pub struct MyNonFungibleToken {
     admin: Address,
     name: Bytes,
     total_count: u32,
-    owners_to_nft_ids: Map<Address, u32>,
+    owners_to_nft_ids: Map<Address, Vec<u32>>,
```
elizabethengelman, line 18: I realized that I made a mistake with the `owners_to_nfts_ids` field and was overwriting the owner's nft id if they had more than one. 🤦‍♀️ 

I added this field so that it would be easier to get all of a user's nfts... but I'm not sure it is worth adding it in here. What do you think @willemneal ?

willemneal: Makes sense here. You could use a` Map<u32, ()>`, turning it into a Set. So adding and removing will be easier and you can use `.keys()` to get the `Vec<u32>`

examples/soroban/nft/src/nft.rs
willemneal: Yep! We should add and `init` decorator! I'll open an issue.

examples/soroban/nft/src/nft.rs
elizabethengelman: Good call! Just pushed up the refactor

crates/loam-cli/examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,126 @@
...
+    subcontract::{IsInitable, IsNonFungible},
+    Contract,
+};
+
+#[contracttype]
+#[derive(IntoKey)]
+pub struct MyNonFungibleToken {
+    admin: Address,
+    name: Bytes,
+    total_count: u32,
```
chadoh, line 17: Why are we making the upper limit so small?? Surely everyone alive will want one of our NFTs! 🙃

crates/loam-cli/examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,126 @@
...
+
+impl IsInitable for MyNonFungibleToken {
+    fn nft_init(&self, admin: Address, name: Bytes) {
+        Contract::admin_get().unwrap().require_auth();
+        MyNonFungibleToken::set_lazy(MyNonFungibleToken::new(admin, name));
+    }
+}
+
+impl IsNonFungible for MyNonFungibleToken {
+    // Mint a new NFT with the given owner address and metadata, returning the id
```
chadoh, line 51: I wonder if we should model copying the comments from the subcontract into the implementation, or leave the comments out in this file. The ones in the `#[subcontract]` are the ones that show up in contract calls.

crates/loam-cli/examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,126 @@
...
+    fn mint(&mut self, owner: Address, metadata: Bytes) -> u32 {
+        owner.require_auth();
```
chadoh, line 53: Our implementation allows anyone to mint themselves NFTs? Seems like most implementations would want to limit this to `Contract::admin_get()`, right?

crates/loam-cli/examples/soroban/nft/src/nft.rs
elizabethengelman: Great point - we can probably leave them out of this file and rely on the ones in the subcontract file 👍

elizabethengelman: [ba7af15](https://github.com/loambuild/loam-sdk/pull/91/commits/ba7af1509a108903b64a29721fcf2f6068cc3ec3)

crates/loam-cli/examples/soroban/nft/src/nft.rs
elizabethengelman: Lol, you're right! Do you think `u128` is too ambitious? 😜

crates/loam-cli/examples/soroban/nft/src/nft.rs
elizabethengelman: @willemneal I also want to make sure that I understand how this works correctly, because after I deployed an update to this contract so that `mint` has `self.admin.require_auth()` I was still able to invoke a mint tx with a non-admin as the source acct and I'm not quite sure why. 

When I invoke mint like this, it fails as expected because `molly` is not the admin account.
```
stellar contract invoke \
--id CDLCRZ6ICRW5KB7RFKKJEJ57JQHXLRS2MIBDZ6B4LAZGLKY2WLAJYYLO \
--source molly \
--network testnet \
-- \
mint \
--owner molly \
--metadata 0a2020...

error: Missing signing key for account GDVVBBT5YPLPGN6K2JTCLDSOCF467O2JPWLB2JCF4MQMX4BFOYFM7NZF (this is the admin address)
```

However, when I mint with `molly` as the source, but with `gene`, who is the admin, as the owner it works. Is the owner also signing the tx and that is why it succeeds?
```
stellar contract invoke \
--id CDLCRZ6ICRW5KB7RFKKJEJ57JQHXLRS2MIBDZ6B4LAZGLKY2WLAJYYLO \
--source molly \
--network testnet \
-- \
mint \
--owner gene \
--metadata 0a20202...

1
```

elizabethengelman:
> I pushed up a couple changes, to update the Cargo.lock as happens when you run just build now. And I also deployed the contract locally to check the help info and found that it was absent, so I fixed that. Check my commit messages for more info.

Looks great, thanks for pushing those changes up!

> Should we add some tests? We could test that the docs I checked manually. Maybe we could also test some common NFT scenarios. We should assume that people will (blindly) copy-paste this into their own project, so we should make sure it works correctly.

I agree! I can get on testing today.

crates/loam-cli/examples/soroban/nft/src/nft.rs
willemneal: This is a shortcoming of the current stellar-cli. It doesn't show you that it signs both the transaction with the source account and the auth entry that is required for `gene`.  If `gene` isn't a global config, then you can try from a different directory with a global key. Then the CLI won't have access to `gene`'s key and won't be able to sign the auth entry. Again this will be make more explicit in the new signing PR.

elizabethengelman: Cool, thanks for that explanation. 🙏  That's what I was thinking, but needed a sanity check to make sure I wasn't missing something in the NFT contract implementation.

crates/loam-cli/examples/soroban/nft/src/nft.rs
```diff
@@ -10,10 +10,10 @@ use crate::subcontract::{IsInitable, IsNonFungible};
...
     admin: Address,
     name: Bytes,
-    total_count: u32,
-    owners_to_nft_ids: Map<Address, Map<u32, ()>>, // the owner's collection
-    nft_ids_to_owners: Map<u32, Address>,
-    nft_ids_to_metadata: Map<u32, Bytes>,
+    total_count: u128,
+    owners_to_nft_ids: Map<Address, Map<u128, ()>>, // the owner's collection
+    nft_ids_to_owners: Map<u128, Address>,
+    nft_ids_to_metadata: Map<u128, Bytes>,
```
willemneal, line 16: Hmm it seems annoying that the metadata is untyped. Would be nice for consumers of the contract to know what the shape of the metadata is. But if is the current standard it's fine, but it's something we should push on to get agreement.

chadoh: I assumed that this was an intentional part of the spec, to allow rapid ecosystem experimentation and _de facto_ standards to arise.

crates/loam-cli/examples/soroban/nft/src/nft.rs
chadoh: Plenty ambitious! It's 10^14 times as many stars [probably](https://www.space.com/26078-how-many-stars-are-there.html) exist in the observable universe. If every one had 10 planets or off-planet-habs and each of those had ten billion people who wanted one of our NFTs, we'd still have 1000x more address space than we needed.

crates/loam-cli/examples/soroban/nft/src/nft.rs
elizabethengelman: Yeah, that's a great point. From what I can tell from [SEP-0039](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0039.md#describing-your-nft) best practices are to include a few key fields in the metadata: name, description, URL, issue, and code. 

What if we start with those fields in a Metadata struct, and then if/when a standard emerges we can update the type? Not sure how difficult it will be to upgrade subcontracts, though with the ability to redeploy, that seems doable.
```
pub struct Metadata {
    name: Bytes,
    description: Bytes,
    url: Bytes,
    issuer: Address,
    code: Bytes,
}
```

alternatively we could have them all be optional fields? 🤔

willemneal: Personally I think the first three should be required. And Perhaps we use `String`?. It's again annoying because the all seem like that is what they will be. So perhaps we just put this forward as our standard. Other contracts will convert from Bytes to string anyway so we are just skipping a step.

chadoh: Could we implement it as two separate subcontracts, `IsSep39` and `IsSep39Extension`, or something like that? Implementing `IsSep39Extension` would override all instances of metadata from `Bytes` to this suggested `Metadata` struct.

chadoh: Or `IsSep39WithTypedMetadata`, or `IsTypedMetadata`, or... ¯\\\_(ツ)_/¯

crates/loam-cli/examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,118 @@
...
+            owners_to_nft_ids: Map::new(env()),
+            nft_ids_to_owners: Map::new(env()),
+            nft_ids_to_metadata: Map::new(env()),
+        }
+    }
+}
+
+impl Default for MyNonFungibleToken {
+    fn default() -> Self {
+        MyNonFungibleToken::new(env().current_contract_address(), Bytes::new(env()))
```
chadoh, line 35: Wait, we are setting the admin to the contract itself?

chadoh: Why even bother implementing `Default`?

chadoh: Is it something to do with the `&mut self` conversation [above](https://github.com/loambuild/loam/pull/91#pullrequestreview-2151758328)?

willemneal: Right now `Default` is required by the macro

crates/loam-cli/examples/soroban/nft/src/nft.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+impl IsInitable for MyNonFungibleToken {
+    fn nft_init(&self, admin: Address, name: Bytes) {
+        self.admin.require_auth();
```
chadoh, line 41: Wait, how is `self.admin` already set? And why allow `admin` to be passed in here, if it's already set? Should `admin` even be part of the `nft_init` params?

chadoh: Having two different kinds of admin—"the one that updates the contract source" vs "the one that mints tokens"—doesn't seem useful?

crates/loam-cli/examples/soroban/nft/src/nft.rs
chadoh: it's a nonsensical default. i don't like it. Will this fix it?

- #92

crates/loam-cli/examples/soroban/nft/src/nft.rs
chadoh: @elizabethengelman here's how @willemneal helped me implement this in EquitX, for now:

- no `admin` attribute on any subcontract other than the `Core` / `Admin` one
- add a [`require_auth` function](https://github.com/EquitXDev/equitxdev/blob/3ac25cd079a0a9831ab6ed12cd94b53d19d417ac/contracts/data-feed/src/lib.rs#L19-L25) to `Contract` 
- while implementing other subcontracts, I can then `use crate::Contract` and [call `Contract::require_auth()`](https://github.com/EquitXDev/equitxdev/blob/3ac25cd079a0a9831ab6ed12cd94b53d19d417ac/contracts/data-feed/src/data_feed.rs#L78)

While the dependencies are sort of circular, this allows one subcontract to make calls into another.

elizabethengelman: awesome, thanks so much!

crates/loam-cli/examples/soroban/nft/src/test.rs
```diff
@@ -0,0 +1,111 @@
...
+fn test_nft() {
+    let (env, client, contract_id) = &init();
+
+    let admin = Address::generate(&env);
+
+    // test admin_get and admin_set
+    match client.admin_get() {
+        Some(_) => assert!(false, "Admin already set"),
+        None => assert!(true),
+    }
```
willemneal, line 36:
```suggestion
    assert!(client.admin_get().is_none(), "Admin already set");
```

crates/loam-cli/examples/soroban/nft/src/test.rs
```diff
@@ -0,0 +1,111 @@
...
+                args: (&admin,).into_val(env),
+                sub_invokes: &[],
+            },
+        }])
+        .admin_set(&admin);
+
+    match client.admin_get() {
+        Some(a) => assert_eq!(a, admin),
+        None => assert!(false, "No admin set"),
+    }
```
willemneal, line 52:
```suggestion
    assert!(matches!(client.admin_get(), Some(admin)), "No admin set");
```

crates/loam-cli/examples/soroban/nft/src/test.rs
```diff
@@ -0,0 +1,263 @@
...
+                fn_name: "transfer",
+                args: (nft_id.clone(), owner.clone(), new_owner.clone()).into_val(env),
+                sub_invokes: &[],
+            },
+        }])
+        .transfer(&nft_id, &owner, &new_owner);
+}
+
+#[test]
+fn test_nft() {
```
elizabethengelman, line 104: While testing with Env, I had trouble getting the tests to pass while running in parallel. ChatGpt told me that the errors could be caused by a "context already borrowed" issue. It kind of feels like the env isn't being reset in between tests, even though I am presumably creating a new env for each test. 

These tests do pass when I run them with just one thread:
```
cargo test --package example-nft --lib -- test --test-threads=1
```

@willemneal have you experienced anything like this while testing with env?

crates/loam-cli/examples/soroban/nft/src/test.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    // try to mint from non-admin
+    let non_admin = Address::generate(env);
+    let metadata = Bytes::from_slice(env, "metadata".as_bytes());
+    mint_nft(env, client, contract_id, &non_admin, &non_admin, &metadata);
+}
+
+#[test]
+#[should_panic]
+#[ignore = "This should panic, but it's not"]
```
elizabethengelman, line 171: I expect that these two tests should panic because we are trying to set/get storage state before the contract is initialized. And I though that it would panic because of this https://github.com/loambuild/loam/blob/bc4cf880b95dc82f9005415c5897abe8b92b61c4/crates/loam-soroban-sdk/src/lib.rs#L29

I could use another set of eyes on this in case I'm missing something.

crates/loam-cli/examples/soroban/nft/src/nft.rs
elizabethengelman: Finally getting back to addressing this. 

I like the idea of creating an `IsSep39Metadata` subcontract, but I'm unsure how to implement it. 

- One way I tried was to have an `IsSep39Metdata` subcontract with a `Metadata` associated type:
```
impl IsTypedMetadata for MyNonFungibleToken {
    type Metadata = MyNonFungibleTokenMetadata;
}
```
The issue I ran into with this is that our `derive_contract` doesn't generate the associate type over to the generated trait `TypedMetadata`. It just adds the `Impl` generated type. I'm not sure if this is something we actually want to add to the macro, or if there is a better way to do this, but I moved on and tried a different idea...

- The other idea was to have `IsTypedMetadata` include a `deserialize_metadata` fn that transforms the `Bytes` metadata into the MyNonFungibleTokenMetadata type. But that is still annoying because the user needs to serialize the metadata into Bytes when minting, etc. 

I'll keep thinking on this, but I could probably use another set of eyes on this.

elizabethengelman: Also, for now I've just added a Metadata type like the following, maybe we can move forward with that and then continue to improve the standard/make sure that the Metadata can be flexible going forward.

```
pub struct Metadata {
    pub(crate) name: String,
    pub(crate) description: String,
    pub(crate) url: String,
}
```

willemneal: With `#[contracttype]` the name of the type exported, so perhaps we could just called it `Sep39Metadata`?

PR #93, opened 2024-07-12
feat: add `derive_contract` to make it macro more explicit

willemneal:
```
#[derive_contract(Core(Admin))
struct Contract;
```
This makes it much clearer which subcontract is being used and which implementation it uses.


chadoh (approved):
let's gooooo

PR #95, opened 2024-07-12
feat: allow creating new projects with loam init

asonntag:
- wraps stellar contract init
- uses copied and modified funtions from stellar contract init to initialize frontend with examples
- however, does not allow `--with-examples` option
- removes hello_world soroban contract created by default with stellar cli

### DIFF
--- a/crates/loam-cli/src/commands/init.rs
+++ b/crates/loam-cli/src/commands/init.rs
@@ -1,0 +1,145 @@
+use soroban_cli::commands::contract::init as soroban_init;
+use std::{
+    fs::{copy, create_dir_all, metadata, read_dir, remove_dir_all, Metadata},
+    io,
+    path::Path,
+};
+
+// command line argument parser
+use clap::Parser;
+
+// TO-DO: Store these somewhere else eventually. Sororban uses examples repo.
+const LOAM_EXAMPLES: &str = "../../../../examples/soroban/";
+const FRONTEND_TEMPLATE: &str = "https://github.com/loambuild/frontend";
+
+/// A command to initialize a new project
+#[derive(Parser, Debug, Clone)]
+pub struct Cmd {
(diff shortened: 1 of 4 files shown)


crates/loam-cli/src/commands/init.rs
```diff
@@ -0,0 +1,145 @@
...
+    fs::{copy, create_dir_all, metadata, read_dir, remove_dir_all, Metadata},
+    io,
+    path::Path,
+};
+
+// command line argument parser
+use clap::Parser;
+
+// TO-DO: Store these somewhere else eventually. Sororban uses examples repo.
+const LOAM_EXAMPLES: &str = "../../../../examples/soroban/";
```
chadoh, line 12: I think this might only work on your machine?

crates/loam-cli/src/commands/init.rs
```diff
@@ -0,0 +1,145 @@
+use soroban_cli::commands::contract::init as soroban_init;
+use std::{
+    fs::{copy, create_dir_all, metadata, read_dir, remove_dir_all, Metadata},
+    io,
+    path::Path,
+};
+
+// command line argument parser
```
chadoh, line 8:
```suggestion
```

chadoh: (no need for this comment)

crates/loam-cli/src/commands/init.rs
```diff
@@ -0,0 +1,145 @@
...
+                );
+                e
+            })?;
+        }
+    }
+
+    Ok(())
+}
+
+// TO-DO: imported from stellar init
```
chadoh, line 86:
```suggestion
// TO-DO: import from stellar init
```

chadoh (approved):
Got it working with `rust-embed`! 🎉

PR #108, opened 2024-07-13
fix(CLI): loam dev clone Arc of build command instead of cloning each time

willemneal:
This makes use of `Arc`, so that just the pointer is cloned and reference count incremented instead of cloning the whole build command each time.


chadoh (approved):
Thanks!

PR #109, opened 2024-07-13
feat(CLI): support `init` scripts for contracts in `environments.toml`

### LINKED ISSUE #74: build-clients: run `init` commands for contracts
Follow-up to #73 

If environments.toml [specifies an `init` script](https://github.com/loambuild/frontend/blob/7df7587b5d3c87d113074d1186eff61bd8ae9c1e/environments.toml#L41-L53) for a contract, then we need to run that script, as the default account, when the contract is first deployed. If the contract has already been deployed (that is, if we already have an alias saved), then we should assume that the `init` script was already run.

Some open design considerations:

1. Maybe we should have a way to run the init script for a specific contract, so people can retry that part of the `build` run if their script is malformed? Not sure what this would look like. As of #58, you can't run `build-clients` directly, it gets run at the end of `build`. So would "just run init for contract-a" functionality get added to `build`? Not sure. Maybe this is too complicated, and we should just tell people to remove the alias for the problem contract, and re-run `loam build`, which is otherwise idempotent.
2. Maybe we should continue with the rest of the `build_clients` logic, even if one of these fails, and mention how to re-run just this `init` logic, based on whatever we decide above?

BlaineHeffron:
Fixes #74 

Runs init scripst which specify invoke commands to be run when first deploying. Checks if there wasn't previously an alias stored to determine when to run the init command. Allows one to specify a source account for the command by prefixing SOROBAN_ACCOUNT=<value> or STELLAR_ACCOUNT=<value>

### DIFF
--- a/crates/loam-cli/src/commands/build/clients.rs
+++ b/crates/loam-cli/src/commands/build/clients.rs
@@ -3,6 +3,7 @@ use crate::commands::build::env_toml;
 use stellar_xdr::curr::Error as xdrError;
 use serde_json;
 use soroban_cli::commands::NetworkRunnable;
+use soroban_cli::commands::network::Args as Network;
 use soroban_cli::utils::contract_hash;
 use soroban_cli::utils::contract_hash;
 use soroban_cli::{commands as cli, CommandParser};
@@ -13,7 +14,6 @@ use serde_json;
 use std::hash::Hash;
 use stellar_xdr::curr::Error as xdrError;
 
-use super::env_toml::Network;
 
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, clap::ValueEnum)]
 pub enum LoamEnv {
@@ -83,7 +83,7 @@ impl Args {
             return Ok(());
         };
 
-        Self::add_network_to_env(&current_env.network)?;
+        Self::add_network_to_env(current_env.network.clone())?;
         Self::handle_accounts(current_env.accounts.as_deref()).await?;
         self.handle_contracts(
             workspace_root,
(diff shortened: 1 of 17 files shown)


BlaineHeffron:
Resolves https://github.com/loambuild/loam-sdk/issues/74

crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -333,4 +348,74 @@ export default new Client.Client({{
...
+
+            eprintln!("  ↳ Executing: soroban contract invoke {}", args.join(" "));
+            let result = cli::contract::invoke::Cmd::parse_arg_vec(&args)?
+                .run_against_rpc_server(None, None)
+                .await?;
+            eprintln!("  ↳ Result: {:?}", result);
+        }
+        eprintln!("✅ Initialization script for {name:?} completed successfully");
+        Ok(())
+    }*/
```
chadoh, line 420: outdated comment?

```suggestion
```

crates/loam-cli/tests/it/build_clients/init_script.rs
```diff
@@ -0,0 +1,42 @@
...
+
+[development.network]
+rpc-url = "http://localhost:8000/rpc"
+network-passphrase = "Standalone Network ; February 2017"
+
+[development.contracts.soroban_token_contract]
+workspace = true
+init = """
+STELLAR_ACCOUNT=bob initialize --symbol ABND --decimal 7 --name abundance --admin bob 
+STELLAR_ACCOUNT=bob mint --amount 2000000 --to bob
```
chadoh, line 21: This is cool! I would also like to see a test without `STELLAR_ACCOUNT`. The `default_account` logic in `build_clients` should mean it works even without this. You're using an implicit default here, so you'll probably need to set `--to alice`, since she's the default.

crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -333,4 +348,74 @@ export default new Client.Client({{
...
+            let mut args = vec!["--id", contract_id];
+            let mut source_account: Option<&str> = None;
+
+            let parts: Vec<&str> = line.split_whitespace().collect();
+            let mut command_parts = vec!["--"];
+
+            for part in &parts {
+                if let Some(value) = part.strip_prefix("STELLAR_ACCOUNT=") {
+                    source_account = Some(value);
+                } else if let Some(value) = part.strip_prefix("SOROBAN_ACCOUNT=") {
```
chadoh, line 373: I don't think the CLI supports `SOROBAN_ACCOUNT` anymore , and I'd rather have less to explain/support/maintain. Should we remove this?

BlaineHeffron: Yeah I think its best to just stick with what the CLI supports. I just had SOROBAN in there because it was in the docs that I was going off.

crates/loam-cli/tests/it/build_clients/init_script.rs
BlaineHeffron: Sure I'll add another test that doesnt use the STELLAR_ACCOUNT directive (I was kind of just combining two tests into one here)

crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -279,6 +281,7 @@ export default new Client.Client({{
 
                 // Check if we have an alias saved for this contract
                 let alias = Self::get_contract_alias(name)?;
+                let is_new_deployment = alias.is_none();
                 if let Some(contract_id) = alias {
```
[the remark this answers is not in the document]
BlaineHeffron: Good catch, I removed the is_new_deployment check since it will skip the init check anyway if the hashes are the same.

BlaineHeffron:
@chadoh comments have been addressed

crates/loam-cli/tests/it/build_clients/init_script.rs
```diff
@@ -0,0 +1,77 @@
...
+        assert!(output.status.success());
+        assert!(String::from_utf8_lossy(&output.stderr)
+            .contains(" -- initialize --symbol ABND --decimal 7 --name abundance --admin alice"));
+        assert!(String::from_utf8_lossy(&output.stderr)
+            .contains(" -- mint --amount 2000000 --to alice"));
+        assert!(String::from_utf8_lossy(&output.stderr).contains(
+            "✅ Initialization script for \"soroban_token_contract\" completed successfully"
+        ));
+        // ensure setting STELLAR_ACCOUNT works
+        env.set_environments_toml(
```
chadoh, line 42: Why not put from here down in a new test?

chadoh (approved):
Looks great. One small test organization question.

PR #110, opened 2024-07-13
chore: ARCHITECTURE.md

willemneal:
This is the first pass at adding an architecture document that outlines how everything fits together.


chadoh (approved):
Great Mermaid chart!

Issue #75, opened 2024-06-20
build-clients: add `contracts.cargo-workspace` shortcut


chadoh:
In our [example `environments.toml` file](https://github.com/loambuild/frontend/blob/7df7587b5d3c87d113074d1186eff61bd8ae9c1e/environments.toml#L26), we say in development and testing environments, you can do this:

```toml
[development.contracts]
# Auto-build, -deploy, -bind, and -import all contracts in Cargo workspace
# Only supported in development & test environments
cargo-workspace = true
```

The initial implementation of the `build-clients` logic in #42 did not support this. You need to explicitly list each of the contracts.

## Design questions

`development.contracts` currently specifies a vector of contracts. This would change one of them to be a special, reserved string, `"cargo-workspace"`. This means you would not be able to have a contract named `cargo-workspace`, which—maybe that's fine? Seems an unlikely contract name. And also, the actual contracts specified here don't work if you use dashes. The generated wasm files use underscores, and you need to match those. But maybe that's a bug? We probably wouldn't want `cargo-workspace` to be the reserved word and `cargo_workspace` to refer to an actual contract.

All this to say: maybe we don't want the reserved word. Maybe instead we:

- Just always assume you want to generate clients for all workspace contracts, if you have an environments.toml file.
- Add a different key, like `development.contracts-from-workspace = true`. 
- A combination of the two. Make `development.contracts-from-workspace` default to `true`, but allow overriding to `false`. Is this unnecessary at this point?

I personally lean toward the first approach. We could add the `contracts-from-workspace` flag later, if we discover situations where it would be useful to have an `environments.toml` file and _not_ build clients for all workspace contracts.

BlaineHeffron:
IMO I like the simplest option here which is to just auto build clients for all workspace contracts. Like you said, if there are use cases for not doing so we can add a flag later.

PR #117, opened 2024-07-16
Feature: automatically build clients for all workspaces 

BlaineHeffron:
Addresses https://github.com/loambuild/loam/issues/75

Note I went with the option of just automatically building them. We could always add an option to toggle this pretty easily if needed. It will still check the names of contracts to ensure there aren't any typos but it uses the package name for the main loop.

### DIFF
--- a/crates/loam-cli/src/commands/build/clients.rs
+++ b/crates/loam-cli/src/commands/build/clients.rs
@@ -279,7 +279,7 @@ export default new Client.Client({{
                 Some(contracts) => contracts.get(&name as &str),
                 None => None,
             };
-            // Skip only if contract is found and its workspace setting is false
+            // Skip only if contract is found and its `client` setting is false
             if let Some(c) = settings {
                 if !c.client {
                     continue;
--- a/crates/loam-cli/src/commands/build/clients.rs
+++ b/crates/loam-cli/src/commands/build/clients.rs
@@ -72,7 +72,11 @@ pub enum Error {
 }
 
 impl Args {
-    pub async fn run(&self, workspace_root: &std::path::Path) -> Result<(), Error> {
+    pub async fn run(
+        &self,
+        workspace_root: &std::path::Path,
+        package_names: Vec<String>,
+    ) -> Result<(), Error> {
         let Some(current_env) =
             env_toml::Environment::get(workspace_root, &self.loam_env(LoamEnv::Production))?
         else {
@@ -84,6 +88,7 @@ impl Args {
         self.handle_contracts(
             workspace_root,
             current_env.contracts.as_ref(),
+            package_names,
             &current_env.network,
         )
         .await?;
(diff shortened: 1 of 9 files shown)


crates/loam-cli/tests/it/build_clients/contracts.rs
```diff
@@ -69,6 +112,9 @@ network-passphrase = "Standalone Network ; February 2017"
 
 [production.contracts]
 hello.workspace = true
+soroban_increment_contract.workspace = false
+soroban_custom_types_contract.workspace = false
+soroban_auth_contract.workspace = false
```
chadoh, line 117: I find it kind of odd to call it `workspace` at this point. Don't you think it looks odd here? We are disabling... the workspace?

Maybe it was always a weird thing to call it. But maybe now we should rename this field to something like `build-client`. Thoughts?

BlaineHeffron:
Looks like some tests are failing, putting as draft for now.

crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -254,94 +259,115 @@ export default new Client.Client({{
...
+        }
+        for name in package_names {
+            let settings = match contracts {
+                Some(contracts) => contracts.get(&name as &str),
+                None => None,
+            };
+            // Skip only if contract is found and its workspace setting is false
+            if let Some(c) = settings {
+                if !c.workspace {
+                    eprintln!("Skipping {name:?} as its workspace setting is false");
```
chadoh, line 285: Haven't looked at the full output here, but my guess is that it will be annoying to see output for a no-op every single time. Do you think it would make sense to just not output anything, if this setting is `false`?

crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -254,94 +259,115 @@ export default new Client.Client({{
...
-                eprintln!("    ↳ contract_id: {contract_id}");
-
-                // Save the alias for future use
-                Self::save_contract_alias(name, &contract_id, network)?;
-
-                // Run init script if we're in development or test environment
-                if self.loam_env(LoamEnv::Production) == "development"
-                    || self.loam_env(LoamEnv::Production) == "testing"
+            } else {
+                eprintln!("Contract {name:?} not found in contracts, but continuing anyway");
```
chadoh, line 289: Having trouble following this new logic. This happens if... you skip explicitly listing the contract in `environments.toml`, but it's part of the Cargo workspace?

If that's true, I don't think we need to say anything here. Saying "not found, but..." makes it sound like an error-state, but this will be the default. The primary happy path. Most people will only see output for this scenario, most of the time.

chadoh (changes requested):
Thanks for doing this so fast! Some design questions.

crates/loam-cli/src/commands/build/clients.rs
BlaineHeffron: Agreed, that was something I meant to take out before the PR

crates/loam-cli/src/commands/build/clients.rs
BlaineHeffron: Sounds good, will remove

crates/loam-cli/tests/it/build_clients/contracts.rs
BlaineHeffron: I changed it to `client` since I cant use `-` in rust variables.

chadoh (changes requested):
You can use kebab-case in `environments.toml` and make it deserialize to snake_case keys. See the `env_toml.rs` definition of `Network` for an example. That said, I think `client = false` looks good. 

A couple remaining mentions of the `workspace` option in here, but otherwise it looks good to me! Any reason it's still marked as draft?

BlaineHeffron:
Nope just forgot to mark ready

chadoh (approved):
Awesome, thanks!

PR #120, opened 2024-07-17
Update the workspace Cargo.toml on loam init

elizabethengelman:


crates/loam-cli/src/commands/init.rs
```diff
@@ -64,10 +67,39 @@ impl Cmd {
...
+        eprintln!("Error reading Cargo.toml file in: {cargo_path:?}");
+        e
+    })?;
+
+    let cargo_toml_str = regex::Regex::new(r#"soroban-sdk = "[^\"]+""#)
+        .unwrap()
+        .replace_all(
+            cargo_toml_str.as_str(),
+            r#"loam-sdk = "0.6.12"
+loam-subcontract-core = "0.7.5""#,
```
elizabethengelman, line 87: Todo:
- [ ] clean up this ugly string
- [ ] fetch the most recent versions of these crates instead of hardcoding them

crates/loam-cli/src/commands/init.rs
```diff
@@ -64,10 +67,39 @@ impl Cmd {
...
 }
 
+// update a soroban project to a loam project
+fn update_workspace_cargo_toml(cargo_path: &Path) -> Result<(), Error> {
+    let cargo_toml_str = read_to_string(&cargo_path).map_err(|e| {
+        eprintln!("Error reading Cargo.toml file in: {cargo_path:?}");
+        e
+    })?;
+
+    let cargo_toml_str = regex::Regex::new(r#"soroban-sdk = "[^\"]+""#)
```
elizabethengelman, line 82: With this approach, of just expecting that the existing Cargo.toml file always has `soroban-sdk` and replacing it, we can probably not worry about https://github.com/loambuild/frontend/pull/5/files

elizabethengelman: Actually, maybe we should remove it from loam's fe template too

elizabethengelman: @chadoh I'm curious what you think about the approach to updating the workspace Cargo.toml file. The ideas I've come up with are:
1. _This is the current approach in this repo_ - assume that the Cargo.toml file has `soroban-sdk` as a workspace dep (since that is what `stellar contract init` is doing), and replace that with `loam-sdk` and `loam-subcontract-core`. I'd like to make one update to make sure to get the most recent released version of `loam-sdk` and `loam-subcontract-core`, but haven't gotten to that just yet.
2. Re-clone the fe template repo, and grab the Cargo.toml from it and use that instead of the existing one. This would probably be the way to make this easiest to update the versions in the future.
3. Get the Cargo.toml file from github's HTTP API... but when we did that in the soroban cli we ran into some rate limiting issues, and I think I'd like to avoid dealing with that again.
4. Embed a Cargo.toml file in the source code, and copy that over. But I'm not sure if that is functionally much different from the first approach.

crates/loam-cli/src/commands/init.rs
chadoh: Is there an easy way to grab it from the cloned frontend repository? That will already have the latest dependencies, and it might be surprising if it mismatches what's generated by `loam init`.

crates/loam-cli/src/commands/init.rs
chadoh: > Re-clone the fe template repo, and grab the Cargo.toml from it and use that instead of the existing one. This would probably be the way to make this easiest to update the versions in the future.

I see, so by the time we get here, we have already discarded the `Cargo.toml` file (and all other files ignored by `soroban_init::Cmd`)? Hence the need to re-clone?

That is kind of a pain!

But still the approach I'm leaning toward 🤔 

Maybe we could send a follow-up PR to Stellar CLI to mark various helper functions `pub`, so we can re-use them here? That might make the re-clone easier, though it still does a re-clone.

Would there be some way to allow us to avoid the re-clone, short of adding functionality that is probably not needed by other frontend templates?

crates/loam-cli/src/commands/init.rs
chadoh: If you're going to do this raw string approach, I think the outdented thing is totes fine. But maybe I'm just used to seeing it; we do it a lot in the guts of the TS Bindings code, and in the `build_clients` tests in this repo.

crates/loam-cli/src/commands/init.rs
chadoh: To be clear, I mean is there some way to allow us to avoid the re-clone _while still grabbing the Cargo.toml from the fe template_, and without doing the GitHub-rate-limit-violating approach described above? Like, does `soroban_init::Cmd` leave the file hanging around somewhere? Can we add some option to make it do so, maybe that isn't exposed in the CLI but is available when using it as a library?

elizabethengelman: > By the time we get here, we have already discarded the Cargo.toml file (and all other files ignored by soroban_init::Cmd)? Hence the need to re-clone?

Exactly. In soroban init we are creating a temporary directory where the FE template is being cloned into, and then we copy all the relevant files into the new project dir. Then at the end of the process, the temporary dir is removed.

> Maybe we could send a follow-up PR to Stellar CLI to mark various helper functions pub, so we can re-use them here? 
I think that the main fn would be this `[clone_repo](https://github.com/stellar/stellar-cli/blob/41a1e0dd0e3e8e34f94b4e78fc7f71d7f15420c7/cmd/soroban-cli/src/commands/contract/init.rs#L269)` fn, which basically puts together a few fns from another crate (gix). I think we could expose that fn publically, but I'm not sure if that makes sense for `stellar-cli` as a whole to expose it, ya know? It doesn't have much to do with stellar stuff.

> Would there be some way to allow us to avoid the re-clone, short of adding functionality that is probably not needed by other frontend templates?
Hm, we could add a flag to the stellar CLI that would tell `stellar contract init` to overwrite the existing Cargo.toml with the one from the FE template. So `stellar contract init --frontend-template <repo-url> --overwrite-workspace-cargo` would overwrite the base workspace Cargo.toml with the FE one. _(maybe we could think of a name for the flag)_. 

Maybe this potentially being useful for other FE templates?🤷‍♀️ 


> does soroban_init::Cmd leave the file hanging around somewhere?
Right now it doesn't, because we're cloning the FE template into a temporary directory that will automatically be deleted when it goes out of scope. I'm not sure if would make sense for any other use case to keep that directory around 🤔 It seems nice that it is temporary and removed once the process is over so the user doesn't have to think about it.

chadoh: What if `stellar contract init --frontend-template <repo-url>` offers to overwrite all conflicting files that are already included by the no-fe-template `stellar contract init`? Rather than just ignoring them? You gotta type `y/n` for each of them. Or you can pipe `yes` into it, `yes | stellar contract init`. And _maybe_ we add a `-y`, or whatever the pattern is, to automatically do that.

That feels like a reasonable addition for every frontend template. It will help frontend template authors understand what's happening, and give them more flexibility.

chadoh (approved):
I think this looks totally fine for our immediate need, though! Let's keep the conversations going and we can add a follow-up issue!

crates/loam-cli/src/commands/init.rs
elizabethengelman: that's an interesting idea! at the moment we're merging the .gitignore and README files - i wonder if we should leave that behavior for the default (when the user chooses no, don't overwrite my file). And then when they choose to overwrite their file, that would skip the merging bit? Or should we do away with the merging all together and rely on the overwriting?

I like how this is helping us shape our frontend template "standard" a bit. Like if we are doing away with the merging of the gitignore and README files, then we should indicate to the template authors that they need to include that stuff in their templates.

elizabethengelman: Also, I just created a rough issue: https://github.com/loambuild/loam/issues/123 that we can take this conversation to

PR #121, opened 2024-07-18
Fix/no account overwrite

BlaineHeffron:
Temporary workaround while we wait for this to be fixed and released: https://github.com/stellar/stellar-cli/issues/1380

This makes it so that `generate` will not be run if the account already exists. I used `keys fund` because its the only command under `keys` that actually hits the network. Presumably if one is testing on a local instance of stellar then their local identity files might be out of date so using `fund` is the only way to guarantee they actually exist on the network that I'm aware of. If there is a better command to query the network for the existence of the key let me know and I will use that.

### DIFF
--- a/crates/loam-cli/src/commands/build/clients.rs
+++ b/crates/loam-cli/src/commands/build/clients.rs
@@ -227,9 +227,12 @@ export default new Client.Client({{
 
     async fn account_exists(account_name: &str) -> Result<bool, Error> {
         // TODO: this is a workaround until generate is changed to not overwrite accounts
-        match cli::keys::fund::Cmd::parse_arg_vec(&[account_name])?.run().await {
-            Ok(_) => Ok(true),
-            Err(_) => Ok(false)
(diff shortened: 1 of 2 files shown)


crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -225,6 +225,17 @@ export default new Client.Client({{
         Ok(())
     }
 
+    async fn account_exists(account_name: &str) -> Result<bool, Error> {
```
willemneal, line 228: You can use `Ok(..is_ok())`

willemneal (changes requested):

chadoh (approved):
If you tried it out and it works as expected, then let's go!

crates/loam-cli/src/commands/build/clients.rs
BlaineHeffron: thanks, updated

BlaineHeffron:
Addressed Willem's comment, should be good to go!

willemneal (approved):

PR #131, opened 2024-07-22
fix: hash check on not yet deployed contracts on local network

elizabethengelman:


crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -170,16 +170,28 @@ impl Args {
...
-        Ok(hex::encode(ctrct_hash) == hash)
+        .await;
+
+        match result {
+            Ok(result) => {
+                let ctrct_hash = contract_hash(&result)?;
+                Ok(hex::encode(ctrct_hash) == hash)
+            }
+            Err(e) => {
+                if e.to_string().contains("Contract not found") {
```
elizabethengelman, line 188: I'm not 100% sure about this, but I think that this one of the issues I was seeing earlier today with `loam dev --build-clients`. 

When running `loan dev --build-clients` I was seeing this error: 
```
Build error: Contract not found: CBNIKD2XXVJTY5H6CRPJ23OCIG5QVC7WEMGN6XDGBERZ6YUV4K6R7R6K
```

And the rest of the `loam dev` operations weren't happening, and the contract clients weren't being built. 

I think this is because I had a new network container running without the contracts deployed. When  `cli::contract::fetch::Cmd` was called it was returning an error, because contract was indeed not deployed yet. But, I think that this is an expected, and okay error for `dev`. So instead of allowing the error to bubble up and stop the `loam dev` command, this change catches that particular error and returns `Ok(false)` which later fns can handle.

chadoh (approved):
Would like @BlaineHeffron to review as well!

BlaineHeffron:
Looks good to me.

PR #133, opened 2024-07-23
feat: add `stellar_asset!` macro. Resolves a token client at compile

willemneal:
Currently you need to hard code the contract ids of existing SAC contracts like "native". This allows this to be resolved at compile time. It does currently require that STELLAR_NETWORK be `local`, `testnet`, `futurenet` or `mainnet` and defaults to `local`.


chadoh:
Once conflicts are resolved it is set to auto-merge.

willemneal:
Oops made a wrong change. You need to add back the soroban-sdk version. Will do tomorrow.

willemneal:
The auto merge triggered before the tests finished which is odd!

Issue #130, opened 2024-07-22
CLI: contract `init` scripts don't understand quotes


chadoh:
Given this commit script for the [equitx data_feed contract](https://github.com/EquitXDev/equitxdev/tree/main/contracts/data-feed):

```toml
[development.contracts.data_feed]
client = true
init = """
admin_set --new-admin me
sep40_init --resolution 300000 --assets "[{\"Stellar\": \"$(stellar contract id asset --asset native)\"} ]" --decimals 14 --base "{\"Stellar\":\"$(stellar contract id asset --asset native)\"}"
"""
```

It should run correctly.

Right now, `loam build` doesn't understand the quotes on the `sep40_init` script the right way. The command works when you run it in your terminal, but not when you run it through `loam build` or `loam dev`.

PR #134, opened 2024-07-24
Fix/support init scripts with quotes

BlaineHeffron:
Fixes https://github.com/loambuild/loam/issues/130

Adds subcommand support and uses `shlex` to parse the arguments so that it properly respects quoted string arguments.

### Limitations

This will not work if you escape quotes inside of quotes. The escaped characters get stripped out by the toml parsing. If you need double quotes in an argument you will need to use single quotes. Or you can escape the escape characters. 

To get the init script in the issue to work use this:
```
[development.contracts.data_feed]
client = true
init = """
admin_set --new-admin me
sep40_init --resolution 300000 --assets '[{"Stellar": "$(stellar contract id asset --asset native)"} ]' --decimals 14 --base '{"Stellar":"$(stellar contract id asset --asset native)"}'
"""
```

Another limitation is that I run the subcommands through `sh` if on linux and `cmd` if on windows. Perhaps we could add an argument to pass the preferred shell. Also, if the subcommand errors, program execution will stop and the error will be printed.

### DIFF
--- a/crates/loam-cli/src/commands/build/clients.rs
+++ b/crates/loam-cli/src/commands/build/clients.rs
@@ -454,7 +454,7 @@ export default new Client.Client({{
             });
 
             // After replace_all, check if there's an error message
-            if error_message != "" {
+            if !error_message.is_empty() {
                 return Err(Error::SubCommandExecutionFailure(
                     failed_subcommand,
                     error_message,
--- a/crates/loam-cli/src/commands/build/clients.rs
+++ b/crates/loam-cli/src/commands/build/clients.rs
@@ -2,7 +2,7 @@
 use crate::commands::build::env_toml;
 use regex::Regex;
 use serde_json;
-use shlex;
+use shlex::split;
 use soroban_cli::commands::NetworkRunnable;
 use soroban_cli::utils::contract_hash;
 use soroban_cli::{commands as cli, CommandParser};
@@ -290,7 +290,7 @@ export default new Client.Client({{
     }
 
     async fn handle_contracts(
-        &self,
+        self,
         workspace_root: &std::path::Path,
         contracts: Option<&Map<Box<str>, env_toml::Contract>>,
         package_names: Vec<String>,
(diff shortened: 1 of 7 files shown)


BlaineHeffron:
One thing I was thinking about is how if the init script fails, it wont be rerun on subsequent `build` commands since the alias for the contract will be present. This makes me think perhaps we should add a --force-init or something of that nature in case the user wants to force it to be run.

chadoh (approved):
I’m impressed! Looks pretty good. Paging @willemneal and @pselle for reviews. 

I think maybe sh/cmd is fine for now? Not sure if adding the option right now is a premature optimization. Ideally, I think we would give the option of which shell to use.

chadoh:
> One thing I was thinking about is how if the init script fails, it wont be rerun on subsequent build commands since the alias for the contract will be present. This makes me think perhaps we should add a --force-init or something of that nature in case the user wants to force it to be run.

yeah, I agree that would be nice. Separate PR for it?

crates/loam-cli/src/commands/build/clients.rs
```diff
@@ -399,38 +406,75 @@ export default new Client.Client({{
...
         name: &str,
         contract_id: &str,
         init_script: &str,
     ) -> Result<(), Error> {
+        let re = Regex::new(r"\$\((.*?)\)").expect("Invalid regex pattern");
+
+        let (shell, flag) = if cfg!(windows) {
+            ("cmd", "/C")
+        } else {
+            ("sh", "-c")
```
willemneal, line 454: Can we expected every unix machine to have `sh`?

BlaineHeffron: `sh` is part of the POSIX standard, which most Unix systems should comply with. I'm sure there might be some old machines that don't have it but I'm not sure we need to worry about that. 

see https://pubs.opengroup.org/onlinepubs/9699919799/
https://www.reddit.com/r/commandline/comments/vc8npd/posix_compliant_shells/

That being said, if the user wants to use a command that would not work in `sh`, I think it makes sense to add an option to specify your shell.

willemneal:
This looks good enough for me. The ideal way to solve this is (and perhaps I can take a look at it since I did it for smartdeploy) to extract the clap logic from stellar-cli, we could also see about exposing it. This way we don't need to use `Command` since we can parse the data and do the invoke directly.

Issue #77, opened 2024-06-20
build-clients: allow specifying accounts using simple strings


chadoh:
In [our example `environments.toml` file](https://github.com/loambuild/frontend/blob/main/environments.toml#L26), we say that we will allow you to set `accounts` in two ways: using the fully-specified object, like `{ name = "alice", default = true }`, or just a string like `"alice"`. 

As of #42, we only support the fully-specified version.

Let's support both. Ideally, `{ name = "alice" }` and `"alice"` would both get deserialized to the same exact thing. If that's too complicated, we may want to do something similar to how the `cargo_toml` package deserializes `dependencies`, [turning them into `Dependency::Simple` or `Dependency::Detailed`](https://gitlab.com/lib.rs/cargo_toml/-/blob/main/src/cargo_toml.rs?ref_type=heads#L1050-1062).

PR #137, opened 2024-07-31
feat: enable string for account definition

BlaineHeffron:
Resolves https://github.com/loambuild/loam/issues/77


chadoh (approved):
Awesome! So clean!

Issue #68, opened 2024-06-13
Use cargo-run-bin, remove soroban-cli install from `just setup`


chadoh:
This project already depends on a certain version of stellar-cli (fka soroban-cli) in (various of) its Cargo.toml file(s). It also installs a pinned version of it using `just setup` (check the Justfile). This `just setup` is what we rely on in CI to get the correct version installed.

Let's instead use [cargo-run-bin](https://github.com/dustinblackman/cargo-run-bin), so we don't need to keep these two versions in-sync.

PR #139, opened 2024-08-01
use cargo-run-bin instead of pinned stellar version

BlaineHeffron:
Uses `cargo-run-bin` crate instead of relying on a pinned stellar version in the justfile (which we'd have to keep synced with the cargo.toml files)

resolves https://github.com/loambuild/loam/issues/68

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ loam-soroban-sdk = { path = "./crates/loam-soroban-sdk" }
 loam-sdk-macro = { path = "./crates/loam-sdk-macro" }
 loam-subcontract-ft = { path = "./crates/loam-subcontract-ft" }
 
+cargo-run-bin = "1.7.3"
 cargo_metadata = "0.18.1"
 thiserror = "1.0.38"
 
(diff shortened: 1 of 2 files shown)


BlaineHeffron:
I ended up having to call a script which I placed in a new directory. Perhaps it should be placed somewhere else / called something different.

stellar_version/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+    for package in metadata.packages {
+        if package_is_using_soroban_cli(&package) {
+            if let Some(version) = get_soroban_cli_version(&package) {
+                println!("{version}");
+                return;
+            }
+        }
+    }
+
+    eprintln!("soroban-cli dependency not found in any crate.");
```
chadoh, line 24: Can/should this now be `stellar-cli` throughout?

```suggestion
    eprintln!("stellar-cli dependency not found in any crate.");
```

chadoh (changes requested):
I think this PR may have gotten somewhat out-of-date since you created it?

chadoh (approved):
✨

chadoh:
Set to auto-merge once merge conflicts are resolved

Issue #150, opened 2024-08-16
Update fungible token subcontract & implementation for latest standard


chadoh:
The [official Soroban FT standard](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0041.md) has evolved since @willemneal first authored `loam-subcontract-ft` and its implementation in the examples folder.

We need to update these to match the latest implementation.

Let's also update the `trait` to include all the comments from the spec, as well as linking to the spec and mentioning that it's SEP41 in the docs for the main `MyFungible` trait.

BlaineHeffron:
Closed via https://github.com/loambuild/loam/pull/155

PR #151, opened 2024-08-19
(WIP) Update token example

pselle:
This works on https://github.com/loambuild/loam/issues/150
- Adds SEP 41 trait to reference
- Some adjustments on Cargo/the contract would not build when I first attempted


chadoh:
@BlaineHeffron this might be useful for you to reference. Closing in favor of #155.

Issue #148, opened 2024-08-16
`loam dev` detects changes but does NOT rebuild


chadoh:
Running `loam dev` while working, I see the following:

<img width="720" alt="image" src="https://github.com/user-attachments/assets/251a1f8e-8c55-4ce7-a869-4a1508501d7d">

It detects every change, but it never runs `build` again.

I also see a non-existent file there, `4913`. No idea what that might be.

Then when I press ctrl-C, before it actually stops the process, it runs `build` one last time.

<img width="1255" alt="image" src="https://github.com/user-attachments/assets/0ec72828-7b15-47f4-b869-e382b27b337e">

PR #152, opened 2024-08-19
fix dev command

BlaineHeffron:
Fixes https://github.com/loambuild/loam/issues/148

There were three issues: the build command was using a Mutex guard unnecessarily since there was already a mutex state to lock when rebuilds occur. This lock was preventing rebuilds and causing the bug.

Second is it was watching for the file `environments.toml` rather than watching the directory  and for the env file within it. This was causing a problem where if there wasn't an `environments.toml` file it would throw an error.

Third issue was the test function that checked for lines from  the stderr stream wasn't actually failing when it should since it wasn't infinitely looping within the timeout method.


crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -142,7 +139,8 @@ impl Cmd {
             .manifest_path
             .parent()
             .unwrap_or_else(|| Path::new("."));
-        let env_toml_path = workspace_root.join("environments.toml");
+        let env_toml_dir = workspace_root;
+        let env_toml_filename = "environments.toml";
```
willemneal, line 143: Why is this a variable? Perhaps move this to the top level as `const ENV_FILE: &str = `

crates/loam-cli/src/commands/dev/mod.rs
```diff
@@ -98,7 +95,7 @@ impl Watcher {
 }
 
 fn is_temporary_file(path: &Path) -> bool {
-    const IGNORED_EXTENSIONS: &[&str] = &["tmp", "swp", "swo"];
+    const IGNORED_EXTENSIONS: &[&str] = &["tmp", "swp", "swo", "swx"];
```
willemneal, line 98: What type of file is this?

BlaineHeffron: my vim editor creates it ```
$ vim --version
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled May 03 2024 02:37:51)
Included patches: 1-579, 1969, 580-1848, 4975, 5016, 5023, 5072, 2068, 1849-1854, 1857, 1855-1857, 1331, 1858, 1858-1859, 1873, 1860-1969, 1992, 1970-1992, 2010, 1993-2068, 2106, 2069-2106, 2108, 2107-2109, 2109-3995, 4563, 4646, 4774, 4895, 4899, 4901, 4919, 213, 1840, 1846-1847, 2110-2112, 2121
```

Comment (chadoh) [reply crates/loam-cli/src/commands/dev/mod.rs:97]: Can we `sw*`?

**Review by chadoh — APPROVED** [review]
Looks like it's mostly working. There are some other issues that I'd like to see fixed, but we can do them as separate issues, if you want.

- need to include `loam dev --build-clients`. `--build-clients` should be default behavior with `dev` (but not with `build`). Honestly, I think it should always be the default, because if you have an `environments.toml` then it seems like you want it, but we can have a vidchat with @willemneal if there's conflicting opinions here.
- output still says it's running _soroban_ commands:
   ```
    Executing: soroban contract invoke --id C…
   ```
   We should update this to `stellar`

**Comment by chadoh** [comment]
Looks like legitimate test failures, too

File: crates/loam-cli/src/commands/dev/mod.rs (hunk 2 above)
Comment (chadoh) [reply crates/loam-cli/src/commands/dev/mod.rs:97]: I'm a little bothered by the fact that we need to explicitly ignore all these temp files. Is there a way to use a system default or user-level setting? Maybe we could use the same settings as [ripgrep's automatic filtering](https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#automatic-filtering)?

Comment (BlaineHeffron) [reply crates/loam-cli/src/commands/dev/mod.rs:97]: It looks like the `ignore` crate might have some default filters. (ignore is what ripgrep uses)

**Comment by BlaineHeffron** [comment]
> Looks like legitimate test failures, too

I'm not sure what the example core stuff is

File: crates/loam-cli/src/commands/dev/mod.rs (hunk 2 above)
Comment (BlaineHeffron) [reply crates/loam-cli/src/commands/dev/mod.rs:97]: Looks like `ripgrep` might use a default extension list to search from rather than excluding stuff: https://github.com/BurntSushi/ripgrep/blob/ea99421ec896fcc9a9cd5367d42a80f89316822f/crates/ignore/src/default_types.rs#L274

Comment (BlaineHeffron) [reply crates/loam-cli/src/commands/dev/mod.rs:97]: I tried using the `ignore` crate's `Gitignore` which is nice because it will default use .gitignore patterns to ignore which we could use to not trigger rebuilds. But you still need to add in stuff like .swp to get it to ignore those, I'm not sure if there is anything built in to ignore those files. 
```
    pub fn new(env_toml_dir: &Path, packages: &[PathBuf]) -> Self {
        let env_toml_dir: Arc<PathBuf> = Arc::new(canonicalize_path(env_toml_dir));
        let packages: Arc<Vec<PathBuf>> = Arc::new(packages.iter().map(|p| canonicalize_path(p)).collect());
        
        let mut builder = GitignoreBuilder::new(&*env_toml_dir);
        for package in packages.iter() {
            builder.add(package);
        }
        let ignores = Arc::new(builder.build().expect("Failed to build GitIgnore"));

        Self {
            env_toml_dir,
            packages,
            ignores,
        }
    }
```

BlaineHeffron: I pushed some code that utilizes any gitignore patterns and I also added the 4913 one. It will still show multiple events for something like a vim save because it is doing multiple operations in the background (including creating a 4913 file which it copies over) but the debouncing takes care of it so that it doesnt rebuild. We could also implement some debouncing on the event handler so that it only prints out the last file modified that it detected which would clean up the console output.

chadoh: That sounds great!

PR #155, opened 2024-09-20
Feat/ft sep41 update

BlaineHeffron:
Updates the fungible token contract and the example implementation to match sep41.


chadoh (approved):
Thanks!

Issue #145, opened 2024-08-02
Investigate: contract not rebinding when quickstart container already has contracts deployed


elizabethengelman:
I have noticed the following scenario, and I think that it's worth looking into:

- When I already have my quickstart container running 
- And have already deployed my loam contracts
- When I try to create a new project sometimes the contracts aren't re-bound to ts clients in the new project

BlaineHeffron:
When you notice it doesn't rebind, do you see if it gets to the `stellar contract bindings typescript` call? Maybe its erroring out before that? Or is it the bindings call thats failing?

elizabethengelman:
Good question - I'm not seeing errors per se, but here are the repro steps. I'm also wondering if this is an edge case that we shouldn't be too concerned about. 🤷‍♀️


Repro steps:

- start docker desktop
- start quickstart `soroban network container start local`
#### new-project
- init a new project: `loam init new-project`
- build: `cd new-project && loam dev --build-clients`
- this works and both example contracts are installed, bound and imported, and the ts files are in `new-project/src/contracts`
```
🌐 using network at http://localhost:8000/rpc
ℹ️ account "me" already exists, skipping key creation
📲 installing "example_core" wasm bytecode on-chain...
    ↳ hash: f491c824a15c7ee3548a5de2b20c987a4a3e6d73aed5082187e5b1de466091e8
🔄 Updating contract "example_core"
🪞 instantiating "example_core" smart contract
    ↳ contract_id: CCOQBAZKJ4DB7C7XB6FNU6FDUC6ICUKEUGAVJ65E4ZLZZKBSEH2FXS7L
🎭 binding "example_core" contract

added 29 packages, and audited 31 packages in 3s

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

> example_core@0.0.0 build
> tsc

🍽️ importing "example_core" contract
📲 installing "example_status_message" wasm bytecode on-chain...
    ↳ hash: 195a674c9770ce3bfcd5c7db42910e23779b8d177c5980b08bf97d04fec341b3
🔄 Updating contract "example_status_message"
🪞 instantiating "example_status_message" smart contract
    ↳ contract_id: CB7KYIO3MU6RDENQHIN4XIXNOLWYQWD5F3SJFCFDFYVAAPLLFBVB6YQW
🎭 binding "example_status_message" contract

added 1 package, and audited 33 packages in 1s

5 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

[...]

PR #156, opened 2024-09-20
Fix/alias issue

BlaineHeffron:
fixes https://github.com/loambuild/loam/issues/145
Also Upgrades to stellar-cli 21.5
The fix is done by creating the `.stellar` directory in the `workspace_root`. This is needed because if that directory doesn't exist within the directory passed to `locator::Args` via `config_dir` it will ignore `config_dir` and fall back to searching up the tree of directories until it finds a `.stellar` directory that already exists.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,7 @@ jobs:
         STELLAR_CONTRACT_ID: core
     services:
       rpc:
-        image: stellar/quickstart:v423-testing
+        image: stellar/quickstart:v438-testing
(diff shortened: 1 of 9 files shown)


chadoh (approved):
Thanks!

chadoh:
What are these pesky RPC tests

chadoh:
We really do need to get these tests passing again. Took a brief stab at it but seeing different perplexing on my own machine rn. Regardless, _this_ PR does not affect the failing tests.

PR #157, opened 2024-09-23
feature: Build clients in the order specified in env file

BlaineHeffron:
This is so that init scripts that may be order dependent will work.


chadoh (approved):
Thank you!

Issue #170, opened 2024-11-05
loam-dev: auto-start local network


chadoh:
### Brief Description

Currently, Loam requires that a Docker container be launched manually before development. To streamline this process, we aim to incorporate the `stellar container` functionality to automatically start a container when using `loam dev`. This will make it easier for new developers to get started and simplify the testing process.

### How to Measure Completion

Completion will be measured by releasing a version of loam-cli that includes the functionality to automatically start a Docker container using the `dev` command.

## Technical Considerations

If https://github.com/stellar/stellar-cli/issues/1701 has not yet been completed, then we should also ensure that `loam dev` has its own functionality to await network health.

PR #176, opened 2024-12-12
add ability to auto start container

BlaineHeffron:
Runs `stellar container network start local` when `run_locally` is set to `true` in `environments.toml`. Resolves https://github.com/loambuild/loam/issues/170


chadoh (approved):
This looks great, thanks! Looks like there's a test failure on `main`. Could you fix that in a follow-up PR?

Issue #172, opened 2024-11-26
Refactor key generation  to use individual key value pairs for variables


BlaineHeffron:
# Loam SDK Storage Setting and Retrieval Refactoring 

This (rough draft) epic outlines tasks to allow setting instance and persistent storage for individual key value pairs, along with setting their TTL in a way that works with the `loam` concept of modifying self for storage values. It is based on a discussion Willem and I had regarding implementation of the `token` example in `soroban-examples`. The current issue is that `IntoKey` will generate a single key value pair in storage (using persistent storage) to store the state of whatever enum you are using. A more efficient means of doing this is to set key value pairs for each piece of data you are storing in the contract. We would also like to still be able to use `Map` types and let `loam` do the work under the hood of creating a unique key for each entry in the map. Currently, the entire `Map` is stored under one ledger entry.

Another issue is that you cannot set instance storage, `IntoKey` only does `persistent`. We would like to add different types to differentiate - e.g `InstanceValue` for instance singletons, `InstanceMap` for key value pairs stored in instance storage, `PersistentMap`, etc. Same for temporary storage.

This would look something like:

```
#[loamtype]
#[derive(Clone)]
pub struct Token {
    admin: InstanceValue<Address>,
    decimal: InstanceValue<u32>,
    name: InstanceValue<Bytes>,
    symbol: InstanceValue<Bytes>,
    balances: PersistentMap<Address, i128>,
    allowances: InstancetMap<(Address, Address), i128>,
}
```

`loamtype` would be used in place of `contracttype` and in the background it would create all of the `DataKey` type enums (see the `token` example in `soroban-examples`) to be used for the generated getters and setters. There could also be SetTTL type methods to put in custom TTL values. 

Another task is to remove the requirement for a default implementation. The new requirement should be that there should be `init` methods if applicable for all subcontracts that can only be called by the `admin`. These should instantiate all instance storage types. A constructor for the contract would call all of the subcontract init methods and it should only be possible to call this constructor once.

## Tasks

- [x] Implement Map type for setting and getting storage
  - [x] Create InstanceValue type
  - [x] Create InstanceMap type
  - [x] Create corresponding Persistent and Temporary types as well
  - [x] Decide on final naming convention

- [x] Implement a macro (e.g `loamtype` ) that enumerates the data keys for all subcontract types in the struct 
  - [x] These would implement getters and setters for all the data keys.

- [x] Remove requirement for default implementation
  - [x] Identify and remove default implementation dependencies
[...]

PR #177, opened 2025-01-09
feat: loam storage

BlaineHeffron:
This accomplishes the goals in https://github.com/loambuild/loam/issues/172

Edit: Now has examples and updated documentation.

Currently all of the soroban examples have been ported to loam but use the `IntoKey` macro. The idea is to use the new storage types here to properly implement them. These will serve as additional examples of how to use the new storage types.


crates/loam-sdk/README.md
```diff
@@ -15,27 +15,23 @@ Build **composable**, **upgradeable**, **secure** Smart Contracts.
 
 # Subcontracts
 
-A subcontract is a type that implements the `IntoKey` trait, which is used for lazily loading and storing the type.
+A subcontract is a type that implements the `loamstorage` trait, which is used for lazily loading and storing the type.
```
willemneal, line 18: `loamstorage` isn't a trait. So perhaps it is a type that implements a subcontract trait, e.g. `IsCore`. Not sure how to phrase this though.

willemneal: Technically implementing a trait with  #[subcontract] requires the Lazy trait.

BlaineHeffron: Yeah, maybe we could change to `lazy` then add, use the `loamstorage` macro along with `PersistentMap`, etc to implement storage. or something along those lines

willemneal: Yeah that sounds good!

willemneal (approved):

PR #178, opened 2025-01-14
fix(docs): Update README.md to binstall for CLI

pselle:
Recommend `cargo binstall` as the method to install Loam CLI


willemneal (approved):
Thank you!

PR #180, opened 2025-01-28
feat: add to_string and log that doesn't require a passed `env`

willemneal:


BlaineHeffron (commented):
Looks good but would be nice to have some tests.

chadoh (approved):

PR #181, opened 2025-01-30
feat: Integrate Smartdeploy as loam registry

willemneal:


contracts/registry/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "loam-registry"
```
pselle, line 2: Is this named `loam-registry` or `Smart Deploy`? The README says SmartDeploy

willemneal: Good point!

PR #191, opened 2025-04-17
fix: copy frontend template for init

BlaineHeffron:
Stellar-cli removed the ability to copy a frontend template for `init` so we need to do that as part of the `init` command.


crates/loam-cli/src/commands/init.rs
```diff
@@ -175,3 +179,37 @@ fn rename_cargo_toml_remove(project: &Path, name: &str) -> Result<(), Error> {
...
+    fs_extra::dir::copy(
+        temp_dir.path(),
+        project_path,
+        &fs_extra::dir::CopyOptions::new()
+            .content_only(true)
+            .overwrite(true),
+    )
+    .map_err(|e| {
+        eprintln!("Error copying frontend files");
+        Error::IoError(io::Error::new(io::ErrorKind::Other, e.to_string()))
```
willemneal: I would just create a new Error for this particular case.  This way the error type can convey that info not a print to stderror.

BlaineHeffron: done

crates/loam-cli/src/commands/init.rs
```diff
@@ -186,14 +190,11 @@ fn clone_repo(repo_url: &str, dest: &Path) -> Result<(), Error> {
         .status()
         .map_err(|e| {
             eprintln!("Error executing git clone");
```
willemneal:
```suggestion
```

crates/loam-cli/src/commands/init.rs
```diff
@@ -208,7 +209,7 @@ fn copy_frontend_files(temp_dir: &TempDir, project_path: &Path) -> Result<(), Er
     )
     .map_err(|e| {
         eprintln!("Error copying frontend files");
```
willemneal:
```suggestion
```

willemneal (approved):

